PR #25, opened 2017-02-16
Bug fix/fc download numbers

Trinkes:
Resolved the ticket #AN-1195


Emadeg (commented):
e7c64764-0a61-4032-bb83-8ea06bf7838a

PR #99, opened 2017-06-05
[AN-1677] - Added field app_installed to PostReviewRequest.java.

FredericoSantinho:


v8engine/src/main/java/cm/aptoide/pt/v8engine/view/dialog/DialogUtils.java
```diff
@@ -58,6 +61,7 @@ public DialogUtils(AptoideAccountManager accountManager, AccountNavigator accoun
     this.bodyInterceptor = bodyInterceptor;
     this.httpClient = httpClient;
     this.converterFactory = converterFactory;
+    installedRepository = RepositoryFactory.getInstalledRepository();
```
Trinkes: inject it

FredericoSantinho: ok

PR #103, opened 2017-06-06
Feature/AN_1660_ShareInAppView

phillipsgoncalves:
Before, in the appview, the share via Spot&Share option would only show up if the app was installed. Now that option is always available and if the app is not installed, a snackbar should be shown with an "install" action.


utilsdesign/src/main/java/cm/aptoide/pt/utils/design/ShowMessage.java
```diff
@@ -210,16 +210,16 @@ public static void asSnack(Fragment fragment, @StringRes int msg) {
     return asSnackObservableInternal(asSnackInternal(fragment, msg));
   }
 
-  public static void asSnack(Activity activity, int msg, int actionMsg,
-      View.OnClickListener action) {
-    Snackbar snackbar = asSnackInternal(activity, msg, actionMsg, action);
+  public static void asSnack(Activity activity, int msg, int actionMsg, View.OnClickListener action,
+      int duration) {
```
Trinkes: duration is not used

PR #111, opened 2017-06-20
Feature/enhance purchase statistics

marcelorbenites:
- New information added to V3 web services interceptor and /payProduct request body.
- Move AdsRepository creation to V8 engine.
- Move method that retrieves application version by package name from AptoideUtils to PackageRepository.


marcelorbenites:
Ticket: AN-1656

PR #113, opened 2017-06-20
Refactor: dataprovider

SithEngineer:
What does this PR do?
This pull-request is the begginning of the refactoring for the dataprovider module.
 
- [x] merged ImageLoader into V8Engine
- [x] merged DataProvider and NetworkClient
- [x] removed lombok from DataProvider
 
Open Aptoide v8 > Navigate and use app


SithEngineer:
This modifications have an higher impact as previously forseen. Due to that, and to the fact that this refactor is not finished, a new pull request will be open later when all impacted code from this modifications are properly versioned and accounted for.

PR #119, opened 2017-06-26
Keeping user image on screen rotation (AN-1613)

SithEngineer:
What does this PR do?
Main objective is to keep the user image on view uppon configuration changes on the device (ex.: screen rotation);
Secondary objective: improve the used Presenter to use the view model and setup the view state when creating the view.


v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
```diff
@@ -22,52 +22,99 @@
...
-        .doOnNext(__ -> navigateBack());
+  @Override public void restoreState(Bundle state) {
+    // does nothing
+  }
 
-    Observable<Void> handleSelectImageClick = view.selectUserImageClick()
-        .retry()
-        .doOnNext(__ -> view.showLoadImageDialog());
+  private void onViewCreatedLoadUserData() {
+    view.getLifecycle()
```
Trinkes: this observable is never unsubscribed

v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
```diff
@@ -22,52 +22,99 @@
...
 
+  private void handleSaveDataClick() {
     view.getLifecycle()
         .filter(event -> event == View.LifecycleEvent.CREATE)
-        .flatMap(
-            __ -> Observable.merge(handleCancelClick, handleSaveDataClick, handleSelectImageClick))
+        .flatMap(__ -> view.saveUserDataButtonClick()
+            .doOnNext(__2 -> view.showProgressDialog())
+            .flatMap(userData -> saveUserData(userData))
+            .retry())
```
Trinkes: why use this retry?

v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
```diff
@@ -22,52 +22,99 @@
...
+        .compose(view.bindUntilEvent(View.LifecycleEvent.DESTROY))
+        .subscribe(__ -> {
+        }, err -> crashReport.log(err));
+  }
+
+  private void handleSelectImageClick() {
+    view.getLifecycle()
+        .filter(event -> event == View.LifecycleEvent.CREATE)
+        .flatMap(__ -> view.selectUserImageClick()
+            .retry()
```
Trinkes: why use this retry?

v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
SithEngineer: on save user data, some error could happen and we need to re-subscribe to the click of UI button that triggers the saving of the data. otherwise in an error with a first click, the button would do nothing on a second click.

v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
SithEngineer: this not necessay and will be removed.

v8engine/src/main/java/cm/aptoide/pt/v8engine/view/account/user/ManageUserPresenter.java
SithEngineer: it is a single, so when one result returns it gets subscribed. will add composition to view destruction even in case that happens before the result though.

PR #120, opened 2017-06-27
Fixed bug in spot and share when username is empty.

FredericoSantinho:


v8engine/src/main/java/cm/aptoide/pt/v8engine/spotandshare/AccountGroupNameProvider.java
```diff
@@ -30,7 +30,7 @@ public AccountGroupNameProvider(AptoideAccountManager accountManager, String man
         .toSingle()
         .flatMap(account -> {
           String username;
-          if (account.isLoggedIn() && (username = account.getNickname()) != null) {
+          if (account.isLoggedIn() && !TextUtils.isEmpty(username = account.getNickname())) {
```
Trinkes: use account.getNickname()!=null && account.getNickname()!="" instead

FredericoSantinho:
Done

PR #126, opened 2017-07-03
[Analytics/AN-1690] - Root install analytics

Trinkes:
Analytic events are now sent when a root install is made informing if there was an error or the installation was successful


FredericoSantinho:
Do you agree with commited changes?

PR #148, opened 2017-07-20
Feature/AN-1782 Validating login on social interactions

SithEngineer:
This PR fixes part of the bug https://aptoide.atlassian.net/browse/AN-1782 regarding the need of checking for user auth before doing social interactions on the Timeline.

It does not validate the secondary case regading the store visibility and the user.


SithEngineer:
using validations to check if user is private and/or store is private

v8engine/src/main/java/cm/aptoide/pt/v8engine/social/presenter/TimelinePresenter.java
```diff
@@ -344,15 +356,58 @@ private void clickOnLikeSocialPost() {
...
+                      return Completable.fromAction(
+                          () -> view.showCreateStoreMessage(SocialAction.LIKE));
+                    } else if (showSetUserOrStoreToPublic(account)) {
+                      return Completable.fromAction(() -> view.showSetUserOrStorePublicMessage());
+                    }
+
+                    final Post post = cardTouchEvent.getCard();
+                    return timeline.like(post, post.getCardId())
+                        .doOnCompleted(() -> view.updatePost(post,
+                            ((AppUpdateCardTouchEvent) cardTouchEvent).getCardPosition()));
```
jdandrade: This throws a class cast exception because the cardTouchEvent might not be from an AppUpdate card. Reason why Likes in Social Posts don't work.

analaragomes:
Use case 2: 
- user logged in
- user is private and does not have store
- presses like/comment/share
- show snack: You need to create a store to %s action: Create
Does not work
Use ana.simoes+271@aptoide.com

analaragomes:
Use case 3:
user logged in
user is private and has a private store
presses like/comment/share
show snack:  You need to set your store or user to public. To change these settings, go to https://www.aptoide.com. action: no action

PR #160, opened 2017-08-01
Bugfix for navigation crash in screenshots

SithEngineer:
An attempt to fix this issue: [AN-1823](https://aptoide.atlassian.net/browse/AN-1823) [fabric](https://www.fabric.io/aptoide2/android/apps/cm.aptoide.pt/issues/59763f4dbe077a4dcc1101fa?time=last-seven-days)

This commit also solves an issue that is not detailed in the ticket but not all screenshots are being displayed. With this PR, this issue is fixed as well.


analaragomes:
I dd not wanted to accept, but too late. Problem: screenshots in appview are not ok. The picture does not have the right aspect ratio. The same does not happen in production.

PR #168, opened 2017-08-02
Share preview dialog checkbox for privacy update

SithEngineer:
What does this pull request do?

This pull request does a minor refactoring in the share dialogs part of the timeline and shows a checkbox in the card the user is sharing for the user to inform us if he wants to become private or stay public (default). This checkbox only appears when the user is "Unlisted".

- [x] Create a user and after inserting a user name, kill the app thus not allowing the app to confirm user visibility preference (private or public) so the user becomes "unlisted"
- [x] Login with an "unlisted" user and try to share a card in timeline

1. create a user and after inserting a user name, kill the app
2. remove the app account in the device settings
3. login with the previous user in the app
4. go to timeline tab
5. press button to share a card but do not press "share" button
6. start web proxy
7. in the popup, a checkbox should be visible. check or un check the box
8. press "share" button
9. in the web proxy a user account sync should have occured before the timeline card share syncing the user access with the selected by the card checkbox
10. select another card and press share. the checkbox should not be visible again

Tickets related to this pull-request: 

- [x] [AN-1783](https://aptoide.atlassian.net/browse/AN-1783)


v8engine/src/main/java/cm/aptoide/pt/v8engine/social/data/share/SharePostViewSetup.java
```diff
@@ -0,0 +1,117 @@
...
+    final String accountStoreName = account.getStore()
+        .getName();
+    final String accountUserNickname = account.getNickname();
+    final String accountUserAvatar = account.getAvatar();
+    final String accountStoreAvatar = account.getStore()
+        .getAvatar();
+    final Account.Access accountUserAccess = account.getAccess();
+
+    storeName.setTextColor(ContextCompat.getColor(context, R.color.black_87_alpha));
+    if (Account.Access.PUBLIC.equals(accountUserAccess)) {
```
jdandrade: This condition should only manage User information visibility (hide/show: username, user avatar). Store stuff should be shown.

PR #213, opened 2017-08-23
Passed all the PNG files trough a compressing algorithm

dfloureiro:
Compressing all the PNG files using "Biu".
The images should not lose quality, at least for the human eye, since it's based on the well known compressing algorithm tinyPNG. Anyway you can check it out yourself.
This procedure should decrease the final apk size in at least 0.6MB (better then nothing 👍).


SithEngineer:
image

Found some files that where badly minified by the used tool.

SithEngineer:
screen shot 2017-10-02 at 16 07 08

PR #216, opened 2017-08-24
Feature/an 1583 home of store share store

FredericoSantinho:


app/src/main/java/cm/aptoide/pt/view/share/ShareStoreHelper.java
```diff
@@ -0,0 +1,39 @@
...
+      if (ShareDialogs.ShareResponse.USING == eResponse) {
+        caseDefaultShare(storeUrl);
+      }
+    }, CrashReport.getInstance()::log);
+  }
+
+  private void caseDefaultShare(String storeUrl) {
+    Intent sharingIntent = new Intent(Intent.ACTION_SEND);
+    sharingIntent.setType("text/plain");
+    sharingIntent.putExtra(Intent.EXTRA_TEXT, storeUrl);
```
jdandrade: missing message:

Text: “Hello! Follow me on Aptoide http://en.aptoide.com/store/<store_name>”

PR #261, opened 2017-09-18
Feature/an 1736 page views

Ribas1:
https://aptoide.atlassian.net/browse/AN-1736


app/src/vanilla/java/cm/aptoide/pt/view/wizard/WizardPageOneFragment.java
```diff
@@ -21,6 +21,11 @@ public static Fragment newInstance() {
     return new WizardPageOneFragment();
   }
 
+  @Override public void onCreate(@Nullable Bundle savedInstanceState) {
```
Trinkes: on SettingsFragment.java the view is registered on onResume method, should this one be registered here?

app/src/main/java/cm/aptoide/pt/analytics/AptoideNavigationTracker.java
```diff
@@ -0,0 +1,61 @@
...
+import cm.aptoide.pt.view.wizard.WizardFragment;
+import java.util.List;
+
+/**
+ * Created by jdandrade on 06/09/2017.
+ */
+
+public class AptoideNavigationTracker implements NavigationTracker {
+
+  public static final String DO_NOT_REGISTER_VIEW = "do_not_register_view";
```
Trinkes: Since NavigationTrackFragment is the one who uses the flag, it should be the one who defines the tag to be used.

PR #275, opened 2017-09-26
[AN-1853] Added an empty state card when there are no posts in a timeline.

jdandrade:
To test:
  Stores > ana2016 > timeline tab > followers > click dsgna/Qualquercoisa > timeline tab > should show the empty state list. This should happen on pull-to-refresh, on retry and on view created


analaragomes:
Button => should only be available if you are the owner of the timeline.

PR #278, opened 2017-09-27
Feature/an 1907 timeline funnel

Trinkes:


analaragomes:
Missing or does not work:
- Event: open_APP => card type=APPS_latest => does not work
- timeline => get app btn => download app => wrong event => previous_context": "HomeFragment" should be previous_context": "TimelineFragment"
- timeline => get app btn => install app => wrong event => previous_context": "HomeFragment" should be previous_context": "TimelineFragment"
- COMMENT_SEND => is not sent
- like event when user is not logged in => is not sent
- open_timeline event - with information of the logged-in user, plus information that we were coming from the login (previous_context) => not sent (only the event was sent not the previous_context)

PR #304, opened 2017-10-12
AN-2026 fixed buttons layout not being aligned in layout bottom

SithEngineer:
Fix for issue https://aptoide.atlassian.net/browse/AN-2026


analaragomes:
The first point of the ticket is working, but not the 2nd one:
" images are not vertically aligned"
This applies for both screens.

PR #307, opened 2017-10-16
AN-2050 partners is broken

analaragomes:
https://aptoide.atlassian.net/browse/AN-2050


dfloureiro:
@analaragomes still broken, does not even build partners... errors in AppViewFragment, SearchWidget and HomeFragment.

PR #311, opened 2017-10-19
AN-2050 Broken Partners build

SithEngineer:
What does this pull request do?

> This pull request resolves the fact that partners flavor was broken due to several changes introduced in vanilla flavor. Code refactoring was also done to mitigate issues related with latest search functionality.
> Out of scope from this PR is an issue found with search that will be dealt later and the customization of views according to partner necessities.

- [x] Checkout code
- [x] Do "gradle clean build" or "./gradlew clean build"

> Checkout code and validate if all flavors are building. Validate if search is working properly in Partners **and** in Vanilla.
> Make sure that code refactoring did not change existing behavior (such as shortcut creation).

Tickets related to this pull-request: 

- [ ] https://aptoide.atlassian.net/browse/AN-2050

Screenshots (if appropriate): N/A

Is there a blog post? No

Does this add new dependencies which need to be added? No


dfloureiro:
When the partner has multistoresearch enabled, there's no all stores button or view.

PR #312, opened 2017-10-19
Feature/an 1977 test fairy

analaragomes:
https://aptoide.atlassian.net/browse/AN-1977


app/build.gradle
```diff
@@ -406,6 +414,9 @@ dependencies {
   compile('com.mopub:mopub-sdk-native-static:4.11.0@aar') {
     transitive = true
   }
+
+  fairyCompile 'testfairy:testfairy-android-sdk:1.+@aar'
```
Trinkes: replace this with exact version

PR #325, opened 2017-10-27
Bugfix/an 2058

paquiquinho:
Analytics fixes for apps timeline: https://aptoide.atlassian.net/browse/AN-2058?src=confmacro
Points 1,2 and 5 aren't working for the comment_send event, but should be working for everything else;
3, 4 and 6 are finished but I couldn't test it for all cards as not all would show up on my timeline to do so. It worked however for the tested situations;


app/src/main/java/cm/aptoide/pt/social/view/TimelineFragment.java
```diff
@@ -614,19 +614,22 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
   }
 
   private void handleSharePreviewAnswer() {
+
+    final ShareEvent[] share = new ShareEvent[1];
     shareDialog.cancels()
-        .doOnNext(shareEvent -> timelineAnalytics.sendShareCompleted(false))
+        .doOnNext(shareEvent -> timelineAnalytics.sendErrorShareCompleted(shareEvent, false, 0))
```
jdandrade: NO ACTION REQUIRED

FYI, as you properly know already, this breaks the clean architecture rules we wanted when implementing the MVP pattern in Timeline. The timeline analytics dependency should only be used in the presenter, the presenter decides what needs to be logged or not, and we try to concentrate that logic there. 

This is not your fault since this method should have been an Observable that would emit the user response (cancel or share) to the presenter and then the presenter should log the analytics events.

app/src/main/java/cm/aptoide/pt/social/view/TimelineFragment.java
```diff
@@ -614,19 +614,22 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
         .doOnNext(event -> sharePostPublishSubject.onNext(event))
-        .doOnNext(shareEvent -> timelineAnalytics.sendShareCompleted(true))
+        .doOnNext(shareEvent -> timelineAnalytics.sendShareCompleted(shareEvent, true))
         .compose(bindUntilEvent(LifecycleEvent.PAUSE))
         .subscribe(shareEvent -> {
+          share[0] =shareEvent;
         }, throwable -> {
           crashReport.log(throwable);
-          timelineAnalytics.sendShareCompleted(false);
+          timelineAnalytics.sendErrorShareCompleted(share[0], false, 1);
```
jdandrade: Is this what we want to log? There should not be errors on the UI when user clicks cancel or share, which is what it looks this is trying to catch.

What we might want to do is follow the sharePostPublishSubject emission and check where is it being subscribed in the TimelinePresenter, and log the error there. This would also clean the 1 element array "hack" that it is used to log errors.

app/src/main/java/cm/aptoide/pt/timeline/TimelineAnalytics.java
```diff
@@ -655,6 +679,11 @@ public void sendCommentCompleted(boolean success) {
 
   public void sendFabClicked() {
     HashMap<String, Object> data = new HashMap<>();
+    final String previousContext = navigationTracker.getPreviousViewName();
+    final String store = navigationTracker.getPreviousScreen().getStore();
```
jdandrade: getPreviousScreen might return null if screen history size is < 2, throwing an NPE

app/src/main/java/cm/aptoide/pt/timeline/TimelineAnalytics.java
```diff
@@ -664,4 +693,203 @@ public void scrollToPosition(int position) {
             bodyInterceptor, httpClient, converterFactory, tokenInvalidator, appId,
             sharedPreferences));
   }
+
+  public HashMap<String, Object> parseEventData(CardTouchEvent event, boolean status, int errorCode){
```
jdandrade: We need to think about a way to shorten this and the parseShareCompletedEventData methods.

Some duplicated code can be removed because there are common parts to all events like:
    data.put("position", event.getPosition());
    data.put("card_type", card.getType());

Maybe we should extract the error parsing switch out to an object that deals with this kind of logic.. on other hand most of this methods (also the others that already existed like sendClickOnPostBodyEvent) make me feel like there is logic that this entity should deal with. 

We should probably talk and try to think of any ideas when you reach this part.

app/src/main/java/cm/aptoide/pt/social/presenter/TimelinePresenter.java
```diff
@@ -755,14 +755,14 @@ private void clickOnLikeSocialPost() {
                     final Post post = cardTouchEvent.getCard();
                     return timeline.like(post, post.getCardId())
                         .andThen(Completable.fromAction(
-                            () -> timelineAnalytics.sendLikeEvent(cardTouchEvent.getPosition(),
+                            () -> timelineAnalytics.sendLikeEvent(cardTouchEvent,
```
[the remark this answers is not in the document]
jdandrade: PS: the boolean is in the LoC below.

jdandrade:
The OPEN_APP event for store cards (STORE/SOCIAL_STORE/AGGREGATED_STORE) is being sent twice, check in charles.

Open timeline > find a store card > open an app

PR #334, opened 2017-11-03
[AN-2075] Apps shown as related apps in articles/social and recomenda…

jdandrade:
…tion cards are prefered to be installed apps.

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/install/PackageRepository.java
+++ b/app/src/main/java/cm/aptoide/pt/install/PackageRepository.java
@@ -130,17 +130,6 @@ public class PackageRepository {
     });
   }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/cm/aptoide/pt/install/PackageRepository.java
```diff
@@ -130,6 +130,17 @@ private void setCachedInstalledPackages(List<PackageInfo> packageInfos) {
     });
   }
 
+  public boolean isPackageNameInstalled(String packageName) {
```
Trinkes: not used

jdandrade: fixed

PR #338, opened 2017-11-07
[StoreMeta] - corrected a bug where app crashes if user profile is open

Trinkes:


app/src/main/java/cm/aptoide/pt/store/view/GridStoreMetaDisplayable.java
```diff
@@ -161,21 +161,26 @@ public boolean hasStore() {
...
+        case GOLD:
+          return GridStoreMetaWidget.HomeMeta.Badge.GOLD;
+        case PLATINUM:
+          return GridStoreMetaWidget.HomeMeta.Badge.PLATINUM;
+        case NONE:
+        default:
+          return GridStoreMetaWidget.HomeMeta.Badge.NONE;
+      }
+    } else {
+      return null;
```
jdandrade: Would it become cleaner if the else return GridStoreMetaWidget.HomeMeta.Badge.NONE and NONE would return a new badge enum value TIN (or DEFAULT)?

app/src/main/java/cm/aptoide/pt/store/view/GridStoreMetaWidget.java
```diff
@@ -139,25 +139,30 @@ public GridStoreMetaWidget(View itemView) {
...
+          break;
+        case GOLD:
+          badgeIcon.setImageResource(R.drawable.gold);
+          break;
+        case PLATINUM:
+          badgeIcon.setImageResource(R.drawable.platinum);
+          break;
+      }
+      badgeIcon.setOnClickListener(v -> badgeDialogFactory.create(badge, storeOwner)
+          .show());
```
jdandrade: Regarding the last comment instead of null checking there would be a case NONE: which would hide the badge and a case DEFAULT: that would show tin.

PR #340, opened 2017-11-08
Feature/an 1581 edit store channels

Ribas1:
Added social channels edition in Edit Store according to following tickets.
- https://aptoide.atlassian.net/browse/AN-1581
- https://aptoide.atlassian.net/browse/AN-2081

Changed a bit the layout so that cancel/save buttons would always stick to the end of screen/scrollview.

To test, check the behaviour of social channels layout, and try to add some channels to a store.


app/src/main/java/cm/aptoide/pt/account/view/store/ManageStoreFragment.java
```diff
@@ -253,6 +443,15 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
     super.onDestroyView();
   }
 
+  @Override public void onDestroy() {
+    super.onDestroy();
+    hideKeyboard();
```
SithEngineer: use onDestroyView to call this and before the super call.

app/src/main/java/cm/aptoide/pt/account/view/store/ManageStoreViewModel.java
```diff
@@ -0,0 +1,133 @@
...
+
+  public void setStoreTheme(StoreTheme storeTheme) {
+    this.storeTheme = storeTheme;
+  }
+
+  public boolean storeExists() {
+    return storeId >= 0L;
+  }
+
+  public List<SimpleSetStoreRequest.StoreLinks> getStoreLinks() {
```
SithEngineer: If the string for each url is all we need why is the WS request body returned?

app/src/main/java/cm/aptoide/pt/account/view/store/ManageStoreFragment.java
```diff
@@ -305,147 +503,195 @@ public void bindViews(View view) {
...
+              .toString())));
+    }
+    if (storeLinksList.isEmpty()) {
+      return Collections.emptyList();
+    }
+    return storeLinksList;
+  }
+
+  private String setYoutubeUrl(String youtubeUsername) {
+    if (youtubeUsername.contains("http")) {
```
SithEngineer: avoid this! use the URL [Pattern matcher from Android](https://developer.android.com/reference/android/util/Patterns.html#WEB_URL)

SithEngineer: again, multiple methods can be avoided by using the enum SocialChannelType

PR #371, opened 2017-12-04
[AN-2148] - Correted wrong instantiation of storeDelLinks in ManageSt…

Ribas1:
…oreViewModel


analaragomes:
This flow does not work:
- login with a user without store
- stores tab
- create store
- set name, theme and picture
- create store is called => ok
- set store is called with store_del_links=null => NOK

PR #389, opened 2017-12-14
Fixed issue AN-2167 regarding generic error message

SithEngineer:
Replaced old specific error string by a generic one. The old specific error is now used in his particular scenario.


Trinkes:
"Unknown error" message is displayed instead of "The image is not valid" on edit user screen

PR #406, opened 2018-01-03
[storeMeta/AN-2170] - solve a bug where the store data isn't refreshe…

Trinkes:
…d after an update from edit screen

Edit store screen.
Store meta were using different navigator instances, because of that store meta was never notified when edit was successful and never update itself.


analaragomes:
Reject:
http://ws75.aptoide.com/api/7/getStoreWidgets/store_id=1244708/context=store
is called with refresh=false and offset=5 
please read the last comment of the ticket

PR #416, opened 2018-01-18
Refactor/an 2155 fc analytics

Trinkes:


app/src/main/java/cm/aptoide/pt/analytics/analytics/AnalyticsManager.java
```diff
@@ -0,0 +1,80 @@
...
+      this.httpKnockEventLogger = httpKnockEventLogger;
+      return this;
+    }
+
+    public AnalyticsManager build() {
+      if (httpKnockEventLogger == null) {
+        throw new IllegalArgumentException("Analytics manager need an okhttp client");
+      }
+      if (eventSenders.size() < 1) {
+        throw new IllegalArgumentException("Analytics manager need at least one logger");
```
jdandrade: eventSenders to logger

app/src/main/java/cm/aptoide/pt/analytics/analytics/RealmEventMapper.java
```diff
@@ -0,0 +1,38 @@
...
+  public RealmEvent map(Event event) throws JsonProcessingException {
+    return new RealmEvent(event.getTimeStamp(), event.getEventName(), event.getAction()
+        .ordinal(), event.getContext(), objectMapper.writeValueAsString(event.getData()));
+  }
+
+  public List<Event> map(List<RealmEvent> realmEvents) throws IOException {
+    ArrayList<Event> events = new ArrayList<>();
+    for (RealmEvent realmEvent : realmEvents) {
+      events.add(new Event(realmEvent.getEventName(),
+          objectMapper.readValue(realmEvent.getData(), new TypeReference<Map<String, Object>>() {
```
jdandrade: question: is it possible that the json is invalid in some weird case? we are writting with the mapper so I don't think it is possible but if it happens it will throw and error and the chain in AptoideBiAnalytics might enter a loop(not sure if it is possible aswell)?

aptoide-database/src/main/java/cm/aptoide/pt/database/realm/RealmEvent.java
```diff
@@ -0,0 +1,49 @@
...
+import io.realm.annotations.PrimaryKey;
+
+/**
+ * Created by trinkes on 12/01/2018.
+ */
+
+public class RealmEvent extends RealmObject {
+  public static String PRIMARY_KEY_NAME = "timestamp";
+
+  @PrimaryKey long timestamp;
```
jdandrade: should be private? any constraint?

PR #424, opened 2018-01-31
An 2233 invalid code facebook event

JDSM01:


analaragomes:
One little detail.... it should be branch: bugfix/AN-2233-Invalid_code_facebook_event instead of branch: AN-2233-Invalid_code_facebook_event :)

PR #426, opened 2018-02-02
Refactor/an 2231 migrate session and first launch analytics

JDSM01:


app/src/main/java/cm/aptoide/pt/analytics/view/AnalyticsActivity.java
```diff
@@ -26,9 +25,7 @@ public boolean is_resumed() {
 
   @Override protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
-    getActivityComponent().inject(this);
-    Analytics.Lifecycle.Activity.onCreate(this);
-
+    analyticsManager = ((AptoideApplication) getApplicationContext()).getAnalyticsManager();
```
jdandrade: Can we inject the AnalyticsManager dependency?

app/src/main/java/cm/aptoide/pt/analytics/view/AnalyticsActivity.java
```diff
@@ -41,18 +38,18 @@ public boolean is_resumed() {
           .getLanguage());
     }
 
-    Analytics.Dimensions.setGmsPresent(AdNetworkUtils.isGooglePlayServicesAvailable(this));
+    ((AptoideApplication) getApplicationContext()).getFirstLaunchAnalytics()
+        .setGmsPresent(AdNetworkUtils.isGooglePlayServicesAvailable(this));
```
jdandrade: Can we inject the FirstLaunchAnalytics dependency?

app/src/main/java/cm/aptoide/pt/ApplicationModule.java
```diff
@@ -948,6 +950,11 @@ public ApplicationModule(AptoideApplication application, String aptoideMd5sum) {
         converterFactory, tokenInvalidator, BuildConfig.APPLICATION_ID, defaultSharedPreferences);
   }
 
+  @Singleton @Provides FirstLaunchAnalytics firstLaunchAnalytics(
```
jdandrade: naming: providesX

app/src/main/java/cm/aptoide/pt/ApplicationModule.java
```diff
@@ -1002,12 +1013,14 @@ public ApplicationModule(AptoideApplication application, String aptoideMd5sum) {
       @Named("aptoideEvents") Collection<String> aptoideEvents,
       @Named("facebookEvents") Collection<String> facebookEvents,
       @Named("fabricEvents") Collection<String> fabricEvents,
-      @Named("flurryEvents") Collection<String> flurryEvents) {
+      @Named("flurryEvents") Collection<String> flurryEvents,
+      @Named("flurry") SessionLogger flurrySessionLogger) {
```
jdandrade: I suggest using @Named("flurrySession") for a clearer understanding and not being confused with the EventLogger instance.

app/src/main/java/cm/aptoide/pt/analytics/view/AnalyticsActivity.java
JDSM01: Fixed. The problem was that it was assuming the class was ActivityResultNavigator with "this", as a result it wasn't injecting the dependecies because the ActivityResultNavigator was already injected.

app/src/main/java/cm/aptoide/pt/analytics/view/AnalyticsActivity.java
JDSM01: Fixed. The problem was that it was assuming the class was ActivityResultNavigator with "this", as a result it wasn't injecting the dependecies because the ActivityResultNavigator was already injected.

PR #439, opened 2018-02-15
[ASV-50] Removed rollback from the app completely;

analaragomes:
This pull-request is supposed to fully remove rollback from the app. I started by removing the Rollback class that managed information persistance in realm and fixed subsequent problems that created until nothing was (hopefully) left;

Testing:
Since rollback was very tied to installations, to test this you should test the installation flows: install, update, and downgrade;

Ticket:
[ASV-50] https://aptoide.atlassian.net/browse/ASV-50


paquiquinho:
The requested changes have been preformed

app/src/main/java/cm/aptoide/pt/AptoideApplication.java
```diff
@@ -494,27 +494,27 @@ public AptoideDownloadManager getDownloadManager() {
...
+                  getDefaultSharedPreferences(), rootAvailabilityManager)).create(this),
+          getRootAvailabilityManager(),
           getDefaultSharedPreferences(),
           SecurePreferencesImplementation.getInstance(getApplicationContext(),
               getDefaultSharedPreferences()),
           RepositoryFactory.getDownloadRepository(getApplicationContext().getApplicationContext()),
           RepositoryFactory.getInstalledRepository(
               getApplicationContext().getApplicationContext()));
-      installManagers.put(installerType, installManager);
+      installManagers.put(InstallerFactory.DEFAULT, installManager);
```
JDSM01: You can remove the use of a SparseArray and only track the InstallManager instance.
This way you can remove the static string from InstallerFactory

PR #443, opened 2018-02-21
[ASV-111] Replace Overflow in App View by Install on TV Icon

paquiquinho:
This pull request aims to remove the overflow from the app view and replace it with an "Install on TV" icon. Note that while the install icon will always display at this point, the overflow will only disappear after the removal of scheduled downloads ([ASV-109]) and Uninstall ([ASV-108]) links is merged with this ticket;

**Where to start**:
fragment_appview.xml;

**Testing**:
Open aptoide -> Open app view -> install on TV icon should be there

**Tickets**:
-Main: ASV-111
-Dependencies: ASV-108, ASV-109


app/src/vanilla/res/menu/fragment_appview.xml
```diff
@@ -36,6 +36,8 @@
 
   <item
       android:id="@+id/menu_remote_install"
+      android:icon="@drawable/ic_download_tv"
+      android:orderInCategory="3"
       android:title="@string/remote_install_menu_title"
-      app:showAsAction="never"/>
+      app:showAsAction="always"/>
```
jdandrade: showAsAction="always" overrides the showAsAction="ifRoom" that the share menu icon has. This can make the share icon go to the overflow before the Install on TV icon, making the item order change. This should only happen now tho with the Schedule download + uninstall option, but shouldn't we make sure that it doesn't?

paquiquinho: I left it this way on purpose so the icon would show. My idea was either forcing "always" on everything, since there are only 3 icons it should be fine, or forcing "ifRoom", this would make everything more scalable and should always show anyway.

PR #479, opened 2018-03-29
[Feature/ASV-229] Alternative Social Flow

jdandrade:
**What does this PR do?**

  This PR adds a new alternative flow for the social Timeline. Now the user will see a recommendation bundle in home with an app. That Recommendation can be an Aptoide recommendation (in-house generated according to user Updates requests in the day before) or a Friend recommendation, which is generated when user clicks install in an app view and selected "Recommend" in the new dialog that is shown. New Analytics events were added to measure this new flow.
  
  This PR also fixes the home fragment tag/screen tracking which were resulting in empty previous_tag on app view's opened from the Home view.

**Database changed?**

   No

- [ ] SocialBundle.java
- [ ] AppViewFragment.java

  Tickets related to this pull-request: [ASV-229](https://aptoide.atlassian.net/browse/ASV-229)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/home/AdMapper.java
```diff
@@ -1,42 +1,56 @@
...
+          .getInfo()
+          .getCpcUrl(), ad.getAd()
+          .getInfo()
+          .getCpdUrl(), ad.getAd()
+          .getInfo()
+          .getCpiUrl(), clickUrl, ad.getAd()
+          .getData()
+          .getName(), ad.getAd()
+          .getData()
+          .getId(), networkId));
```
JDSM01: ad.getAd() is used several times, shouldn't it be extracted?

app/src/main/java/cm/aptoide/pt/home/BundlesResponseMapper.java
```diff
@@ -41,11 +44,31 @@
...
+            if (card.hasUser()) {
+              appBundles.add(
+                  new SocialBundle(applicationsToApps(apps, type, widget.getTag()), type, event,
+                      widget.getTag(), card.getUser()
+                      .getAvatar(), card.getUser()
+                      .getName()));
+            } else {
+              appBundles.add(
+                  new SocialBundle(applicationsToApps(apps, type, widget.getTag()), type, event,
+                      widget.getTag(), R.mipmap.ic_launcher, marketName));
```
JDSM01: widget.getTag, widget.getViewObject and widget.getTitle is being called a lot, maybe it could be set in the start of the for.

app/src/main/java/cm/aptoide/pt/home/FakeBundleDataSource.java
```diff
@@ -46,11 +46,11 @@
 
     List<Application> tmp1 = new ArrayList<>();
     FeatureGraphicApplication aptoideFeatureGraphic =
-        new FeatureGraphicApplication("Aptoide", icon, 0, 1000, "cm.aptoide.pt", 300, icon);
+        new FeatureGraphicApplication("Aptoide", icon, 0, 1000, "cm.aptoide.pt", 300, icon, "", "");
```
JDSM01: Repeatable values, maybe static final variables?

app/src/main/java/cm/aptoide/pt/home/FakeBundleDataSource.java
```diff
@@ -71,7 +71,8 @@
     appBundles.add(appBundle);
     appBundles.add(appBundle1);
     appBundles.add(appBundle2);
-    appBundles.add(new AdBundle("Highlighted", Collections.emptyList(), null, ""));
+    appBundles.add(
+        new AdBundle("Highlighted", new AdsTagWrapper(Collections.emptyList(), ""), null, ""));
```
JDSM01: Should this be here?

app/src/main/java/cm/aptoide/pt/home/Home.java
```diff
@@ -29,4 +33,17 @@ public Home(BundlesRepository bundlesRepository) {
   public boolean hasMore() {
     return bundlesRepository.hasMore();
   }
+
+  @NonNull private List<HomeBundle> addFakeSocialBundleTo(HomeBundlesModel homeBundlesModel) {
```
JDSM01: Not being called

app/src/main/java/cm/aptoide/pt/home/AdMapper.java
jdandrade: extracted

app/src/main/java/cm/aptoide/pt/home/FakeBundleDataSource.java
jdandrade: In normal code, I would say yes. But I think in this fake bundle source, that is only used for testing, gets more flexibility and readability by not reusing some of its values. In this example there are also terrible names like tmp and tmp1 but they are strictly used for testing but I will rename them

app/src/main/java/cm/aptoide/pt/home/FakeBundleDataSource.java
jdandrade: Just wanted to make sure that an AdBundle was included in the fake bundles list for the test, even tho it doesn't have any ads inside. The test is not verifying if ads are shown in the UI tho. I think we should make this test later, also fill this bundle with Ad objects.

app/src/main/java/cm/aptoide/pt/home/Home.java
jdandrade: moved the unused logic to the fake bundle data source.

app/src/main/java/cm/aptoide/pt/ApplicationModule.java
```diff
@@ -1107,8 +1108,14 @@ public ApplicationModule(AptoideApplication application, String aptoideMd5sum) {
     return new AdMapper();
   }
 
-  @Singleton @Provides BundlesResponseMapper providesBundlesMapper() {
-    return new BundlesResponseMapper();
+  @Singleton @Provides HomeAnalytics providesHomeAnalytics(AnalyticsManager analyticsManager,
+      NavigationTracker navigationTracker) {
+    return new HomeAnalytics(navigationTracker, analyticsManager);
+  }
+
```
JDSM01: Doesn't it belong in FragmentModule instead?

PR #483, opened 2018-04-04
[ASV-225] clear history bn

JDSM01:
**What does this PR do?**

This PR aims to put the back navigation the same as the youtube app. 
This means 
1) When you press an item in the BN and then press back you should go to the last item of the BN or leave the app
Example1: Home -> AppView (pressing on an app) -> Home -> (Back) -> Leave App
Example2: Home -> AppView (pressing on an app) -> Stores -> (Back) -> Home
2) It should keep only one instance of the fragments
Example: Home -> Stores -> App -> Stores -> App -> Stores -> (Back) -> App -> (Back) -> Home -> (Back) -> Leave app
3) When you navigate to view from the BN without using the BN and you press back it should ignore the fact that it is on a BN view
Example: Home -> Stores -> A store View -> (Search in Stores and go for result) -> SearchView -> (Back) -> A StoreView

**Database changed?**

 No

BottomNavigationNavigator.java

  Test the different flows examplified above and other flows that might have been missed.

  ASV-225

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [x] Partners build
- [x] Unit tests pass


app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,153 @@
...
+    this.defaultStoreName = defaultStoreName;
+    this.bottomNavigationAnalytics = bottomNavigationAnalytics;
+    this.searchAnalytics = searchAnalytics;
+    bottomNavigationItens = new ArrayList<>();
+    this.fragmentNavigator = fragmentNavigator;
+  }
+
+  public void navigateToBottomNavigationItem(int bottomNavigationPosition) {
+    switch (bottomNavigationPosition) {
+      case 0:
```
phillipsgoncalves: These attributes related to the position could be extracted to a field and reused on each of the navigateTo methods.
This way, it will be easier to change the "positions" in the future, without having to change it on more than one place.

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,153 @@
...
+
+public class BottomNavigationNavigator {
+
+  private final static String EVENT_ACTION =
+      "https://ws75.aptoide.com/api/7/getStoreWidgets/store_id=15/context=stores";
+  private final FragmentNavigator fragmentNavigator;
+  private final String defaultStoreName;
+  private final BottomNavigationAnalytics bottomNavigationAnalytics;
+  private final SearchAnalytics searchAnalytics;
+  private ArrayList<Integer> bottomNavigationItens;
```
phillipsgoncalves: Same typo as before. Itens -> Items

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,153 @@
...
+    bottomNavigationItens.remove(0);
+    navigateBackToBottomNavigationItem(bottomNavigationPosition);
+  }
+
+  public boolean canNavigateBack() {
+    return bottomNavigationItens.size() > 1;
+  }
+
+  private void checkAndReplaceItem(int newItem) {
+    for (int i = 0; i < bottomNavigationItens.size(); i++) {
```
phillipsgoncalves: This is dangerous. You should be carefull while removing items from a list while iterating through it. It can cause a Concurrent Modification Exception.

app/src/main/java/cm/aptoide/pt/home/BottomNavigationActivity.java
```diff
@@ -31,22 +21,24 @@
...
   private Animation animationdown;
 
   @Override protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-    getActivityComponent().inject(this);
-
     setContentView(LAYOUT);
     navigationSubject = PublishSubject.create();
     bottomNavigationView = (BottomNavigationView) findViewById(R.id.bottom_navigation);
+    super.onCreate(savedInstanceState);
```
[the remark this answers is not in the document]
JDSM01: if a fragment needs to do something with the activity onCreate it will break if it has it below. I can change it back, because right now it has no use having it as it is

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
JDSM01: fixed. although, I can change it to work with BottomNavigationItems if you think that's a better solution, that way we can use the other get(Object). For now it goes with that solution as I don't have more time today

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,161 @@
...
+    int bottomNavigationPosition = bottomNavigationItems.get(1);
+    bottomNavigationItems.remove(0);
+    navigateBackToBottomNavigationItem(bottomNavigationPosition);
+  }
+
+  public boolean canNavigateBack() {
+    return bottomNavigationItems.size() > 1;
+  }
+
+  private void checkAndReplaceItem(int newItem) {
```
phillipsgoncalves: This method name could be more generic

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,161 @@
...
+        fragment =
+            MyStoresFragment.newInstance(getStoreEvent(), "default", "stores", StoreContext.home);
+        position = 2;
+        break;
+      case appsPosition:
+        fragment = new AppsFragment();
+        position = 3;
+        break;
+    }
+    if (fragment != null && position != -1) {
```
phillipsgoncalves: This should be inside the default case the switch statement

app/src/main/java/cm/aptoide/pt/home/BottomNavigationMapper.java
```diff
@@ -31,24 +31,46 @@ public BottomNavigationItem mapItemClicked(Integer menuItemId) {
...
-        bottomNavigationId = 2;
+        bottomNavigationPosition = 2;
         break;
       case APPS:
-        bottomNavigationId = 3;
+        bottomNavigationPosition = 3;
         break;
     }
-    if (bottomNavigationId == -1) {
+    if (bottomNavigationPosition == -1) {
```
phillipsgoncalves: This should be on the default case of the switch statement

app/src/main/java/cm/aptoide/pt/home/BottomNavigationMapper.java
```diff
@@ -31,24 +31,46 @@ public BottomNavigationItem mapItemClicked(Integer menuItemId) {
...
+        bottomNavigationPosition = 1;
+        break;
+      case R.id.action_stores:
+        bottomNavigationPosition = 2;
+        break;
+      case R.id.action_apps:
+        bottomNavigationPosition = 3;
+        break;
+    }
+    if (bottomNavigationPosition == -1) {
```
phillipsgoncalves: Same as above, this should be on the default case of the switch statement

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
phillipsgoncalves: Could use bottomNavigationItems.indexOf to find the listPosition variable that is being calculated here

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,161 @@
...
+        break;
+    }
+  }
+
+  public void navigateToHome(BottomHomeFragment bottomHomeFragment) {
+    navigateToSelectedFragment(0, bottomHomeFragment);
+  }
+
+  public void navigateToSearch(SearchResultFragment searchResultFragment) {
+    navigateToSelectedFragment(1, searchResultFragment);
```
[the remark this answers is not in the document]
JDSM01: Fixed

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,161 @@
...
+      case appsPosition:
+        bottomNavigationAnalytics.sendNavigateToAppsClickEvent();
+        AppsFragment appsFragment = new AppsFragment();
+        navigateToApps(appsFragment);
+        break;
+    }
+  }
+
+  public void navigateToHome(BottomHomeFragment bottomHomeFragment) {
+    navigateToSelectedFragment(0, bottomHomeFragment);
```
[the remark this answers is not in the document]
JDSM01: All fixed

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
```diff
@@ -0,0 +1,161 @@
...
+  public void navigateToStore(MyStoresFragment myStoresFragment) {
+    navigateToSelectedFragment(2, myStoresFragment);
+  }
+
+  public void navigateToApps(AppsFragment appsFragment) {
+    navigateToSelectedFragment(3, appsFragment);
+  }
+
+  public void navigateBack() {
+    int bottomNavigationPosition = bottomNavigationItems.get(1);
```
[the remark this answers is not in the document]
JDSM01: You're right but for 2 reasons I think it's better not to change it, we would have to teste all again and this way I think it's a bit more clear on what it's being done (we're accessing the previous fragment and removing the current one)

app/src/main/java/cm/aptoide/pt/home/BottomNavigationNavigator.java
JDSM01: updateBottomNavigationItemsList

PR #566, opened 2018-05-24
[ASV-541] StoreFragment darker color

JDSM01:
**What does this PR do?**

   This PR removes the darker line between the toolbar and tabs

store_activity.xml

Go to a store and check if it doesn't have the dark line now

https://aptoide.atlassian.net/browse/ASV-541

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass

### DIFF
--- a/app/src/main/res/layout/store_activity.xml
+++ b/app/src/main/res/layout/store_activity.xml
@@ -35,6 +35,7 @@
         android:layout_width="match_parent"
         android:layout_height="@dimen/tabsHeight"
         android:background="?attr/backgroundTabs"
+        android:elevation="4dp"
         android:theme="?attr/toolbarTheme"
         />
 


app/src/main/res/layout/store_activity.xml
```diff
@@ -35,6 +35,7 @@
         android:layout_width="match_parent"
         android:layout_height="@dimen/tabsHeight"
         android:background="?attr/backgroundTabs"
+        android:elevation="4dp"
```
jdandrade: after talking with Francisco about this and testing in a device, I think it would be a better solution to remove both this and toolbar's 4dp to have them both at the same level. It doesn't look like it is affecting anything in other screens. I think it is a better approach that could save some confusion in the future if we have to mess with elevations again.

PR #591, opened 2018-06-08
[ASV-568] Apps in BN click feedback

phillipsgoncalves:
**What does this PR do?**

   This aims to improve the interaction with the updates in the BN section. Before, every time a user clicked a button it would take a long time to react. As a temporary solution we decided to set the progress of these apps to indeterminate, until the downloads lib provides us the new download state. This was the solution for every update click.

There is still room for performance improvement.
I have spent a few time researching about the low performance and we think that it is related with Realm and the amount of insertions on the database. Every time there is a progress update or a download status change, there is a database insertion. According to my measurements most of the insertions are taking more than 100ms.

**Database changed?**

 No

- [ ] AppsManager.java

  Open the apps section in the BN and try to update all apps, or update one individual app. Pause, resume and cancel the update. There should be indeterminate progress to give feedback to the user.

  Tickets related to this pull-request: [ASV-568](https://aptoide.atlassian.net/browse/ASV-568)

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/home/apps/AppsAdapter.java
```diff
@@ -321,4 +321,27 @@ public void setAvailableUpdatesList(List<App> availableUpdates) {
...
+    int indexOfApp = listOfApps.indexOf(app);
+    if (indexOfApp != -1) {
+      App application = listOfApps.get(indexOfApp);
+      if (application.getType() == App.Type.UPDATE) {
+        setIndeterminate(indexOfApp, (UpdateApp) application);
+      }
+    }
+  }
+
+  private void setIndeterminate(int indexOfApp, UpdateApp application) {
```
JDSM01: naming: Maybe setIndeterminateStatus

app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -289,8 +289,10 @@ private boolean mapIndeterminate(Download download) {
     Install.InstallationStatus status = Install.InstallationStatus.UNINSTALLED;
     if (download != null) {
       switch (download.getOverallDownloadStatus()) {
-        case Download.FILE_MISSING:
         case Download.INVALID_STATUS:
+          status = Install.InstallationStatus.INITIAL_STATE;
+          break;
```
JDSM01: Question: why install status -> initial state in case of a download invalid status?

phillipsgoncalves: Invalid status is the status that is provided when the download was just initialized. We are mapping that to the install correct status which is INITIAL_STATE.

phillipsgoncalves:
Regarding the 1st, it is what was happening before on the old version, so for simplicity we decided to keep the same behavior, even though i don't agree with it and i think it should be set to its initial state. I hope to change it in the future, but i can not do it in this ticket as this one was just regarding the feedback of the play, pause, cancel, resume and update all buttons.

Regarding the 2nd, we added the indeterminate progress bar status because the buttons feedback was too slow. Every time you pressed any action (pause cancel or resume) it was taking too long (the reason for this is being investigated, it looks like it is related to Realm - there is a spike for this already). From what i understood about your explanation, it is setting to the indeterminate status after pressing the pause button (which is the expected behavior to give feedback to the user). Also, in the update all action, only one download is taken care of at a time, that is why all the other ones are on the "indeterminate" status, until their time arrives (they are in the IN_QUEUE status). I assume this was the reported problem.

app/src/main/java/cm/aptoide/pt/home/apps/AppsAdapter.java
phillipsgoncalves: Don't know if status would be correct term since the status that is being set is the STANDBY. Also, this attribute is being used to set the indeterminate on the progress bar, whose method is also called setIndeterminate, that is why i decided to name it the same way.

PR #620, opened 2018-07-02
[hotfix/ASV-733] OOM on glide's load

JDSM01:
**What does this PR do?**

   Implements a possible solution to solve the Out of Memory problem with glide's load.
   In glide 3.x the default diskcachestdrategy was Resource/Result and in glide 4.x the default is Automatica which may be the cause of that.

**Database changed?**

No

ImageLoader.java

 I wasn't able to replicate this, fabric says that this happens mostly on android 4.x, probably due to these being older and having less memory.
 Just go to places where the glide is being used to be sure that the diskCacheStrategy didn't affect other stuff

https://aptoide.atlassian.net/browse/ASV-733

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/networking/image/ImageLoader.java
+++ b/app/src/main/java/cm/aptoide/pt/networking/image/ImageLoader.java
@@ -18,6 +18,7 @@ import android.view.WindowManager;
 import android.widget.ImageView;
 import cm.aptoide.pt.utils.AptoideUtils;
 import com.bumptech.glide.Glide;
+import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import com.bumptech.glide.load.resource.bitmap.CenterCrop;
 import com.bumptech.glide.request.FutureTarget;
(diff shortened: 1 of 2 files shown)


app/src/main/java/cm/aptoide/pt/networking/image/ImageLoader.java
```diff
@@ -123,13 +122,14 @@ public static void cancel(Context context, View target) {
         placeHolderDrawableId);
   }
 
-  public Target<Drawable> loadWithCircleTransformAndPlaceHolder(String url, ImageView imageView,
+  private Target<Drawable> loadWithCircleTransformAndPlaceHolder(String url, ImageView imageView,
```
jdandrade: Should this be private even though it is only used in this class? The ImageLoader goal is to work like a lib, so in my opinion its contract should be public for anyone using it. Otherwise to use it it would require to change the ImageLoader dependency.

PR #660, opened 2018-08-07
[bugfix/ASV-859] similar and in app message layout after install

JDSM01:
**What does this PR do?**

  It removes the similar apps header if there's no similar apps
  It removes the apps messages when the is downloading
  It puts the similar apps layout below the appc message layout

**Database changed?**

   No

NewAppViewFragment.java

Go to an app with in-app purchase with appcoins (2048) press install and check if the appc message disappears.
Comment the code of the mapToSimilar in NewAppViewFragment so that it returns an empty list. Go to an app with in-app purchase with appcoins (2048), scroll down to check if there's the header of the similar apps (there shouldn't be one). press install and check if similar apps appear (it shouldn't).

https://aptoide.atlassian.net/browse/ASV-859

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/vanilla/java/cm/aptoide/pt/app/view/NewAppViewFragment.java
```diff
@@ -1232,13 +1242,14 @@ private SpannableString formatAppCoinsRewardMessage() {
     if (model.isDownloading()) {
       downloadInfoLayout.setVisibility(View.VISIBLE);
       install.setVisibility(View.GONE);
-      similarBottomView.setVisibility(View.GONE);
-      similarDownloadView.setVisibility(View.VISIBLE);
       showSimilarDownload = true;
+      appcInfoView.hideInfo();
```
jdandrade: Probably hide/show should be done when we listen to the click on install/cancel and not when we show the download, we are mixing download ui logic with app related info.

PR #668, opened 2018-08-17
ASV-877 - Fixed appview app icon flicker and over scroll bug

franciscoaleixo:
**What does this PR do?**
The goal is to fix the known app icon flicker on app view. It also prevents the action bar from getting stuck when flinging upwards. The downside of this fix is that no shadow is present in the action bar, I couldn't find a fix for that.

**Database changed?**

No

build.gradle. The only issue that can arise is with incrementing the support lib version.

Check that if when scrolling on appview, the app icon does not flicker and that you are able to fling upwards without the action bar getting stuck.

Additionally, incrementing the support lib version can break other things that I can't anticipate.

[ASV-877](https://aptoide.atlassian.net/browse/ASV-877)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


franciscoaleixo:
Search action icon should be fixed now. Removed collapseActionView flag as it was causing the bug and actually conflicts with the 'always' tag anyway, so it wasn't having any effect.

phillipsgoncalves:
Accepted but waiting to do a separate release

PR #684, opened 2018-08-29
[ASV-972] Deleted removeAll in the onSharedPreferencesChanged method,…

tiagopedrinho:
… to stop from wrongly removing all the updates from the db, before resetting them.

**What does this PR do?**

   Tries to fix a bug where the excluded updates were being cleared when changing beta versions, HW filter and system apps settings.
 
**Database changed?**

   No

- [ ] SettingsFragment.java

  Go to updates, exclude some updates with long click. Then so to settings, change any of these filters (HW, beta versions, system apps). Then in settings so to excluded updates. The previously excluded updates should be there.

ASV-972

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


jdandrade:
Pull request is ready to be accepted but I need to recheck the reasons why we were resetting the excluded updates when those specific settings changed. We might be missing some business requirement here that was implemented back then.

PR #708, opened 2018-09-17
[Feature/ASV-994] Reactions custom views + interactor/network model

jdandrade:
**What does this PR do?**

   This adds the possibility of showing reactions inside vanilla. They are currently not being shown anywhere. 

   You should look for 2 different parts in this PR. The custom views (entry point is ReactionsPopup.java) and the reactions model (entry point being the Reactions.java).

   The reactions custom views are done by having a custom linearlayout (ReactionsView) that has 5 reaction objects in it. Each reaction object is rendered by lottie using LottieAnimationView and added to ReactionsView. Appropriate callbacks were done in order to return information about what reaction was clicked. Check ReactionsPopup Documentation to see if it is clear.

   2nd phase was the Reactions interactor that at the moment just has an interface to set a reaction with an id and a reactiontype (might be subject of change in the future). The SetReactionsRequest loadDataFromNetwork is throwing an exception on purpose because there is no webservice available yet to do this. Thats also the reason why there is a ReactionFakeService - so we can continue the implementation without the web dependency.

  ReactionsPopup show() currently only supports api19+. This will need to be handled when we show reactions in the future. 

**Database changed?**

   No

- [ ] ReactionsPopup.java
- [ ] Reactions.java

   To check the reactions custom view, just show it in any button of the application. For instance, I've tested it in the rate button in the appview and then shown the reactions view.

  Tickets related to this pull-request: [ASV-994](https://aptoide.atlassian.net/browse/ASV-994)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/reactions/ui/ReactionsView.java
```diff
@@ -0,0 +1,90 @@
...
+        new ViewGroup.LayoutParams(WIDTH_REACTIONS, HEIGHT_REACTIONS);
+
+    for (Reaction reaction : reactions) {
+      reaction.setReactionParams(reactionParams);
+      this.addView(reaction.getView());
+    }
+  }
+
+  public int getReactionsViewHeight() {
+    return 8 + 8 + HEIGHT_REACTIONS;
```
paquiquinho: why 8+8? Is this related to padding/margins? If so, isn't it dangerous?

jdandrade: Extracted to constant & injected to reaction.java
This will probably change when when we make the view integration, since showAsDropDown has to go.

app/src/main/java/cm/aptoide/pt/reactions/ui/ReactionsPopup.java
```diff
@@ -0,0 +1,142 @@
...
+    });
+  }
+
+  /**
+   * Show the reactions popup anchored to the view specified during construction.
+   *
+   * @see #dismiss()
+   */
+  public void show() {
+    popup.showAsDropDown(anchorView, (int) anchorView.getX(),
```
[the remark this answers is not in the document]
jdandrade: Depending on UI/UX - this will have to change when we do the view integration, I think it is better to change then.

jdandrade:
Fixed the animation problem & clicks outside the popup. The problem was somehow related to the background of the PopupWindow - which if not set it stays null and has side effects in some versions.

PR #729, opened 2018-10-09
[ASV-1068] AppNext AB Test

franciscoaleixo:
**What does this PR do?**

  Adds an AB Test experiment for the ad that appears on similar apps inside an appview. The first case is the default (old) one, where the ad is retrieved from our webservices, and it accounts for 90% of the cases. The other case is the new appnext ad (10%).

**Database changed?**

   No

- [ ] AppViewManager.java
- [ ] AdsRepository.java
- [ ] SimilarAdExperiment.java

  Check if each case is working normally
  1st case (90%): Old ad should show up in similar apps
  2nd case (10%): New ad should show up in similar apps instead.

The test case the device gets attributed to can be changed by manually rewriting the response in Charles, or by changing the variable value in debug.

**Note:** APPNEXT_PLACEMENT_ID needs to be set in the global gradle.properties. The value can be found [here](https://aptoide.atlassian.net/browse/ASV-1057).

  Tickets related to this pull-request: [ASV-1068](https://aptoide.atlassian.net/browse/ASV-1068)

**Other notes:**

   The way this experiment is setup is a bit different from the other active ones, in the sense that it's not using a Boolean to return the experiment assigment, but rather directly returning the relevant value to the chain.

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/abtesting/experiments/SimilarAdExperiment.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.List;
+import rx.Scheduler;
+import rx.Single;
+
+/**
+ * Created by franciscoaleixo on 08/10/2018.
+ */
+
+public class SimilarAdExperiment {
+  private final String AD_IDENTIFIER = "ASV-1068-AppNext_Test";
```
paquiquinho: Maybe call this EXPERIMENT_ID like in other experiments? I think it makes more sense;

PR #730, opened 2018-10-11
[feature/ASV-1089] Install_PopUp

JDSM01:
**What does this PR do?**

Made the necessary changes in order to launch the install popUp in the end
The cachePath was also added to the initial path of the download. This led me to delete some stuff from the oldAppViewFragment in order to not have to add new dependencies to an already not used class

**Database changed?**

   No

-AptoideDownloadManager.java

  Make a download and see if it launches the install pop up

https://aptoide.atlassian.net/browse/ASV-1089

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


phillipsgoncalves:
Also, while doing some downloads through the Apps section (different method to create the download in the download factory), of file not found on the moveCompletedDownloadFiles method

downloadmanager/src/main/java/cm/aptoide/pt/downloadmanager/AptoideDownloadManager.java
```diff
@@ -168,6 +167,15 @@ public void start() {
         .toCompletable();
   }
 
+  @Override public void moveCompletedDownloadFiles(Download download) {
```
phillipsgoncalves: This method should be on install manager instead of download manager.

downloadmanager/src/main/java/cm/aptoide/pt/downloadmanager/DownloadManager.java
```diff
@@ -37,4 +37,6 @@
   Completable removeDownload(String md5);
 
   Completable invalidateDatabase();
+
+  void moveCompletedDownloadFiles(Download download);
```
phillipsgoncalves: This method should be on install manager.

app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -562,4 +574,37 @@ public boolean wasAppEverInstalled(String packageName) {
...
+  public void moveCompletedDownloadFiles(Download download) {
+    for (final FileToDownload fileToDownload : download.getFilesToDownload()) {
+      Logger.getInstance()
+          .d("AptoideDownloadManager", "trying to move file : "
+              + fileToDownload.getFileName()
+              + " "
+              + fileToDownload.getPackageName());
+      if (FileUtils.fileExists(fileToDownload.getFilePath())) {
+        String newFilePath = getFilePathFromFileType(fileToDownload);
+        fileUtils.copyFile(cachePath, newFilePath, fileToDownload.getFileName());
```
phillipsgoncalves: The copy should be from the folder where the file is already (so from fileToDownload.getFilePath())
Also, to avoid moving the same file from one folder to the same, we should make a verification if it is different than the newPath

PR #742, opened 2018-10-18
[Bugfix/ASV-1119] - Similar apps impressions are now registered on install click

franciscoaleixo:
**What does this PR do?**

   This fixes the problem where if a user were to open appview and install without scrolling, ad impression would not register.

**Database changed?**

   No

- [ ] Foo.java
- [ ] Bar.kt

  [ASV-1119](https://aptoide.atlassian.net/browse/ASV-1119)

  [ASV-1119](https://aptoide.atlassian.net/browse/ASV-1119)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


franciscoaleixo:
As far as I know, impression should only be recorded once per view/per ad. It's just meant to inform that the user has seen the ad, there's probably not much added benefit to repeat the event. But @analaragomes might know better about this.

PR #748, opened 2018-10-26
[Feature/ASV-1106] AppNext Highlighted Ad

franciscoaleixo:
**What does this PR do?**

   The first ad in the highlighted bundle in Home is now from AppNext following an AB Test. Metrics for this are also implemented. This is well defined in the ticket linked below.

**Database changed?**

   No

- [ ] HighlightedAdExperiment.java
- [ ] HomePresenter.java

  Test with Charles if A/B test is working correctly, there's two buckets:
     - "appnext_ad": the first app in the highlighted bundle is an appnext ad (20%)
     - "no_appnext_ad": there is no AppNext ad in the first position (80%)
   
   For the "appnext_ad" bucket, impression and action analytics should be tested with Charles, as well impression and action events sent to wasabi.

  [ASV-1106](https://aptoide.atlassian.net/browse/ASV-1106)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -86,6 +98,27 @@ public HomePresenter(HomeView view, Home home, Scheduler viewScheduler, CrashRep
...
+        .filter(event -> event.equals(View.LifecycleEvent.CREATE))
+        .flatMap(__ -> adsManager.appNextAdClick()
+            .observeOn(Schedulers.io())
+            .flatMap(result -> {
+              AppNextNativeAd ad = result.getAd();
+              homeAnalytics.sendAdClickEvent(ad.getStars(), ad.getPackageName(), 0,
+                  "ads-highlighted", HomeEvent.Type.AD, ApplicationAd.Network.APPNEXT);
+              return highlightedAdExperiment.recordAdClick();
+            }))
+        .retry()
```
jdandrade: this retry is on the lifecycle event, if it fails it will not wait for an appnext ad click unless a lifecycle event occurs before.

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -160,6 +193,8 @@ public HomePresenter(HomeView view, Home home, Scheduler viewScheduler, CrashRep
         .observeOn(viewScheduler)
         .doOnNext(created -> view.showLoading())
         .flatMapSingle(created -> loadBundles())
+        .filter(bundlesModel -> !hasAppNextHighlightedAd(bundlesModel))
+        .flatMapSingle(__ -> loadAppNextAd("ads-highlighted"))
```
jdandrade: In the future we need to change this. The loadbundles should emit each time a new bundle loads. (highlighted, appnext, recommends) and not block the view while waiting.
This should be only a quick fix now since it requires a lot of effort doing the refactor at the moment.

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -30,10 +37,13 @@
   private final AdMapper adMapper;
   private final AptoideAccountManager accountManager;
   private final HomeAnalytics homeAnalytics;
+  private final HighlightedAdExperiment highlightedAdExperiment;
```
jdandrade: This should probably be in the home interactor (home.java) and not the presenter.

PR #759, opened 2018-11-07
[ASV-1139] - Added version code user property

franciscoaleixo:
**What does this PR do?**

   Adds new user property: version code, so we can track changes between vercodes.

**Database changed?**

   No

- [ ] AptoideApplicationAnalytics.java

  See if the user property "version code" is being sent to analytics.

  [ASV-1139](https://aptoide.atlassian.net/browse/ASV-1139)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/AptoideApplication.java
+++ b/app/src/main/java/cm/aptoide/pt/AptoideApplication.java
@@ -336,6 +336,7 @@ public abstract class AptoideApplication extends Application {
     getRootInstallationRetryHandler().start();
     AptoideApplicationAnalytics aptoideApplicationAnalytics = new AptoideApplicationAnalytics();
     aptoideApplicationAnalytics.setPackageDimension(getPackageName());
+    aptoideApplicationAnalytics.setVersionCodeDimension(getVersionCode());
     accountManager.accountStatus()
         .map(account -> account.isLoggedIn())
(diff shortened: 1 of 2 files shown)


app/src/main/java/cm/aptoide/pt/AptoideApplication.java
```diff
@@ -941,6 +942,16 @@ public ReadPostsPersistence getReadPostsPersistence() {
     return readPostsPersistence;
   }
 
+  public String getVersionCode() {
+    String version = null;
+    try {
+      PackageInfo pInfo = getPackageManager().getPackageInfo(getPackageName(), 0);
+      version = pInfo.versionName;
```
jdandrade: bug here, should be vercode

app/src/main/java/cm/aptoide/pt/AptoideApplication.java
```diff
@@ -941,6 +942,16 @@ public ReadPostsPersistence getReadPostsPersistence() {
...
+    } catch (PackageManager.NameNotFoundException e) {
```
jdandrade: Can we return something like "NaN" - This should never happen but that way we could easily track the exception.

PR #786, opened 2018-11-21
[monetisation/ASV-1187] IronSource Interstitial

franciscoaleixo:
**What does this PR do?**

   Implements IronSource interstitials that should appear when the user clicks Install on AppView. Per specification, it is under an AB test (90% of the users will see it, while 10% will not). Analytics for ad impressions and clicks are also implemented. A user property is also added ("ab_test_ironsource") to help analyze data.

**Note:** IronSourceAdRepository is not being injected into AdsManager because IronSource needs access activity lifecycle, which means it cannot be provided in ApplicationModule (and AdsManager needs to be in ApplicationModule).

**Database changed?**

   No

- [ ] IronSourceInterstitialAdExperiment.java
- [ ] IronSourceAdRepository.java

  Test if interstitials are retrieved and shown with the AB test working (check Charles), as well as the analytics being sent. Also check if the facebook user property "ab_test_ironsource" is being set.

  [ASV-1187](https://aptoide.atlassian.net/browse/ASV-1187)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


analaragomes:
use "network" instead of "Network".
use "IronSource" instead of "ironSource"
Missing:
[App_View_Interact], data = [{Action=tap on app

app/src/main/java/cm/aptoide/pt/AptoideApplicationAnalytics.java
```diff
@@ -30,4 +30,12 @@ public void setVersionCodeDimension(String versionCode) {
...
         .d("Facebook Analytics: ", response.toString()));
   }
+
+  public void setIronSourceAbTestGroup(boolean isControlGroup) {
+    Bundle bundle = new Bundle();
+    bundle.putString("ab_test_ironsource",
+        isControlGroup ? "a_without_ironsource" : "b_with_ironsource");
+    AppEventsLogger.updateUserProperties(bundle, response -> Logger.getInstance()
+        .d("Facebook Analytics: ", response.toString()));
+  }
```
jdandrade: I would move this to its own analytics, like IronSourceAnalytics or even ABTesting analytics. AptoideApplicationAnalytics was created for logging events of stuff that is processed in the Application class.

app/src/main/java/cm/aptoide/pt/AptoideApplicationAnalytics.java
```diff
@@ -30,4 +30,12 @@ public void setVersionCodeDimension(String versionCode) {
     AppEventsLogger.updateUserProperties(bundle, response -> Logger.getInstance()
         .d("Facebook Analytics: ", response.toString()));
   }
+
+  public void setIronSourceAbTestGroup(boolean isControlGroup) {
+    Bundle bundle = new Bundle();
+    bundle.putString("ab_test_ironsource",
+        isControlGroup ? "a_without_ironsource" : "b_with_ironsource");
+    AppEventsLogger.updateUserProperties(bundle, response -> Logger.getInstance()
```
jdandrade: Won't this only be updated if the user goes to the appview? If yes won't this bring problems to the analysis?

app/src/main/java/cm/aptoide/pt/abtesting/experiments/IronSourceInterstitialAdExperiment.java
```diff
@@ -0,0 +1,77 @@
...
+      AptoideApplicationAnalytics aptoideApplicationAnalytics) {
+    this.abTestManager = abTestManager;
+    this.scheduler = scheduler;
+    this.ironSourceAdRepository = ironSourceAdRepository;
+    this.aptoideApplicationAnalytics = aptoideApplicationAnalytics;
+  }
+
+  public Observable<Experiment> loadInterstitial() {
+    return abTestManager.getExperiment(EXPERIMENT_ID)
+        .observeOn(scheduler)
```
jdandrade: We should be careful with this because we are forcing the mainthread but we are possibly not only touching the view after this (calls to analytics for instance). Don't think we should quick fix this since it is an ab testing experiment.

app/src/main/java/cm/aptoide/pt/ads/IronSourceAdRepository.java
```diff
@@ -0,0 +1,87 @@
...
+    IronSource.setInterstitialListener(new InterstitialListener() {
+      @Override public void onInterstitialAdReady() {
+        if (showInterstitial) {
+          showInterstitialAd();
+          showInterstitial = false;
+        }
+      }
+
+      @Override public void onInterstitialAdLoadFailed(IronSourceError ironSourceError) {
+        Log.i(TAG, "Interstitial Ad failed to load. Reason: " + ironSourceError.getErrorMessage());
```
jdandrade: Should not be using Log but Logger.getInstance() since it has logic to disable in release and to be controlled by the Toolbox app.

app/src/main/java/cm/aptoide/pt/ads/IronSourceAdRepository.java
```diff
@@ -0,0 +1,87 @@
...
+
+      @Override public void onInterstitialAdClosed() {
+      }
+
+      @Override public void onInterstitialAdShowSucceeded() {
+        impressionSubject.onNext(null);
+      }
+
+      @Override public void onInterstitialAdShowFailed(IronSourceError ironSourceError) {
+        Log.i(TAG, "Interstitial Ad failed to show. Reason: " + ironSourceError.getErrorMessage());
```
jdandrade: Logger.getInstance()

app/src/main/java/cm/aptoide/pt/ads/IronSourceAdRepository.java
```diff
@@ -0,0 +1,87 @@
...
+import com.ironsource.mediationsdk.IronSource;
+import com.ironsource.mediationsdk.logger.IronSourceError;
+import com.ironsource.mediationsdk.sdk.InterstitialListener;
+import rx.subjects.PublishSubject;
+
+public class IronSourceAdRepository {
+  private static final String TAG = "IronSourceAdRepository";
+  private final Activity activity;
+  private final PublishSubject<Void> impressionSubject;
+  private final PublishSubject<Void> clickSubject;
```
jdandrade: Should avoid multiple subjects - we could use only 1 with an Event object that would have information about whether it is an impression or a click

app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -775,7 +812,11 @@ private void handleInstallButtonClick() {
                                   appViewModel.getPackageName(), appViewModel.getDeveloper()
                                       .getName(), action.toString()))
                               .doOnCompleted(() -> showRecommendsDialog(account.isLoggedIn(),
-                                  appViewModel.getPackageName())));
+                                  appViewModel.getPackageName()))
+                              .toSingleDefault(true)
+                              .delay(3, TimeUnit.SECONDS)
```
jdandrade: Was this discussed with Felix? time based approach seems to be really bad and inconsistent in terms of ux. Also it looks like we are throwing this on top of the share dialog.

app/src/main/java/cm/aptoide/pt/view/MainActivity.java
```diff
@@ -48,6 +50,8 @@
     setupUpdatesNotification();
 
     attachPresenter(presenter);
+
+    ironSourceAdRepository.initialize();
```
jdandrade: attachPresenter should be the last method called here. Don't want to risk having something not initialized being used in the presenter (even tho it is not the case)

app/src/main/java/cm/aptoide/pt/app/AppViewManager.java
```diff
@@ -157,6 +166,26 @@ public AppViewManager(InstallManager installManager, DownloadFactory downloadFac
...
 
+  public Observable<Experiment> initializeInterstitialAd() {
+    return ironSourceInterstitialAdExperiment.loadInterstitial();
+  }
+
+  public Single<Experiment> showInterstitialAd() {
+    return ironSourceInterstitialAdExperiment.showInterstitial();
+  }
+
+  public Observable<Boolean> recordInterstitialImpression() {
```
jdandrade: Not being used.

PR #821, opened 2018-12-17
[ASV-1260] Claim dialog navigation now produces a result to propagate…

paquiquinho:
**What does this PR do?**

   This PR aims to update the app status in the promotions dialog in the case of a successful claim;

**Database changed?**

No

- [ ] PromotionsNavigator.java

Test the claim flow and make sure that when a claim is successfully made, the promotions view updates it's status. Test both the wallet and one of the other apps as these are being handled differently;

  Tickets related to this pull-request: [ASV-1260](<https://aptoide.atlassian.net/browse/ASV-1260>)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogFragment.java
+++ b/app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogFragment.java
@@ -21,7 +21,6 @@ import android.widget.ImageView;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 import cm.aptoide.pt.R;
-import cm.aptoide.pt.navigator.FragmentNavigator;
 import cm.aptoide.pt.networking.IdsRepository;
 import cm.aptoide.pt.networking.image.ImageLoader;
 import cm.aptoide.pt.utils.AptoideUtils;
@@ -52,7 +51,7 @@ public class ClaimPromotionDialogFragment extends DialogFragment
   @Inject ClaimPromotionsManager claimPromotionsManager;
   @Inject IdsRepository idsRepository;
   @Inject PromotionsAnalytics promotionsAnalytics;
-  @Inject FragmentNavigator fragmentNavigator;
+  @Inject ClaimPromotionsNavigator navigator;
   private ClipboardManager clipboard;
   private ClaimPromotionDialogPresenter presenter;
   private ProgressBar loading;
@@ -170,8 +169,7 @@ public class ClaimPromotionDialogFragment extends DialogFragment
     genericErrorView = view.findViewById(R.id.generic_error);
 
     presenter = new ClaimPromotionDialogPresenter(this, new CompositeSubscription(),
-        AndroidSchedulers.mainThread(), claimPromotionsManager, promotionsAnalytics,
-        fragmentNavigator);
+        AndroidSchedulers.mainThread(), claimPromotionsManager, promotionsAnalytics, navigator);
     presenter.present();
     handleRestoreViewState(savedInstanceState);
   }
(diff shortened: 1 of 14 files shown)


app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogFragment.java
```diff
@@ -51,6 +52,7 @@
   @Inject ClaimPromotionsManager claimPromotionsManager;
   @Inject IdsRepository idsRepository;
   @Inject PromotionsAnalytics promotionsAnalytics;
+  @Inject FragmentNavigator fragmentNavigator;
```
jdandrade: I would inject a Navigator specific for this view, for the sake of coherence with other views.

app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogFragment.java
```diff
@@ -327,6 +330,18 @@ public void onResume() {
     this.dismiss();
   }
 
+  @Override public Observable<ClaimDialogResultWrapper> dismissGenericMessage() {
```
jdandrade: This method name is strange, shouldn't be tied to the dismiss logic since this is just a click listener with a mapper.
Now that I look into it the old method was probably what led to this - should we change both?

app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogPresenter.java
```diff
@@ -154,6 +162,17 @@ private void handleCaptchaCancelClick() {
         }));
   }
 
+  private void handleDismissGenericMessage() {
+    subscriptions.add(view.dismissGenericMessage()
+        .doOnNext(message -> {
+          fragmentNavigator.popDialogWithResult(new Result(PromotionsNavigator.REQUEST_CODE,
+              message.isOk() ? Activity.RESULT_OK : Activity.RESULT_CANCELED,
+              new Intent().setPackage(message.getPackageName())));
```
jdandrade: this logic should be in the navigator.

app/src/main/java/cm/aptoide/pt/promotions/PromotionViewApp.java
```diff
@@ -106,6 +106,10 @@ public boolean isClaimed() {
     return isClaimed;
   }
 
+  public void setClaimed() {
+    isClaimed = true;
+  }
```
jdandrade: We should usually make our data objects immutable to make them as predictable as possible. I don't think we should change this now since we have time constraints and the error was not made in this PR - but we should change this if we have opportunity in the future.

app/src/main/java/cm/aptoide/pt/promotions/PromotionsNavigator.java
```diff
@@ -1,22 +1,36 @@
 package cm.aptoide.pt.promotions;
 
+import android.app.Activity;
 import android.os.Bundle;
 import cm.aptoide.pt.navigator.FragmentNavigator;
+import rx.Observable;
 
 public class PromotionsNavigator {
+  static final int REQUEST_CODE = 6666;
```
jdandrade: I think this constant should have a more specific name like CLAIM_REQUEST_CODE

app/src/main/java/cm/aptoide/pt/promotions/ClaimPromotionDialogFragment.java
paquiquinho: fixed

app/src/main/java/cm/aptoide/pt/promotions/PromotionViewApp.java
paquiquinho: agreed, this was done because of time

PR #832, opened 2018-12-26
[ASV-1267] Promotions events (reload and icon impression)

JDSM01:
**What does this PR do?**

 Adds 2 events related to promotions
-When you’re in the captcha dialog and you press reload an event with action reload should be sent
-When you go to home, if there are promotions an event with action impression should be sent
(Check ticket below for more detail on ticket information)

**Database changed?**

No

HomeAnalytics.java
PromotionAnalytics.java
ClaimPromotionDialogPresenter.java

Go to home and check if the impression event is sent on the logicat
Go to the promotions fragment and install the wallet and go trough the claim flow. When you reach the captcha, press reload and check if the event appears on the logicat

https://aptoide.atlassian.net/browse/ASV-1267

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/promotions/PromotionsAnalytics.java
```diff
@@ -132,4 +133,14 @@ public void sendPromotionsAppInteractInstallEvent(String packageName, float appc
           PROMOTIONS_INTERACT, AnalyticsManager.Action.CLICK, getViewName(true));
     }
   }
+
+  public void sendRefreshCaptchaEvent(String packageName) {
+    final Map<String, Object> data = new HashMap<>();
+    data.put(ACTION, "reload");
```
phillipsgoncalves: I would extract this action "reload" as well and in order for it to be just like the others that are already extracted

app/src/main/java/cm/aptoide/pt/home/HomeAnalytics.java
```diff
@@ -27,6 +27,10 @@
   static final String TAP_ON_CARD_DISMISS = "tap on card dismiss";
   static final String TAP = "tap";
   static final String VIEW_CARD = "view card";
+  private static final String ACTION = "action";
+  private static final String BUNDLE_TAG = "bundle_tag";
+  private static final String PROMOTION_ICON = "promotion-icon";
```
phillipsgoncalves: On the ticket it was written "promotions-icon". I know you just extracted the variable that was being used before. This might not be a problem, if Ana correctly analyses it on her side.

JDSM01: Changed to promotion-icon

PR #841, opened 2019-01-04
[ASV-1215] Configuration values from gradle

phillipsgoncalves:
**What does this PR do?**

 This PR aims at adding certain required values to the gradle properties so that they can be set from the BO while trying to produce a cobrand apk.
This values correspond to:
Key: APTOIDE_THEME
value (default): default
Key: COBRAND_OEMID
value (default): 0
Key: COBRAND_MARKET_NAME
value (default): cobrand

While trying to inject the theme and the market name from the gradle properties file, I noticed that:
1. Most of the views are not using the right theme, they are using the aptoide default theme in hardcoded way. This is already reported on the ticket: [ASV-1314](https://aptoide.atlassian.net/browse/ASV-1314)

2. The market name was harcoded on most of the views. I extracted this variable to the ApplicationModule and it is now being correctly injected on the places where it is used. 

To set the cobrand oemid on the downloads link only for the cobrand flavour, i had to extract the code that manages the paths and create an oemid provider that will return an empty string on the vanilla flavour.

**Database changed?**

   No

- [ ] DownloadFactory.java

Open the app both in vanilla and cobrand flavours and check:
If the theme is being correctly applied (change the theme on gradle properties for another color so that you can see the changes)

  Tickets related to this pull-request: [ASV-1215](https://aptoide.atlassian.net/browse/ASV-1215)

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle
- [ ] Partners build
- [ ] Unit tests pass


app/src/main/java/cm/aptoide/pt/view/BaseActivity.java
```diff
@@ -30,7 +30,7 @@ public ActivityComponent getActivityComponent() {
           .plus(aptoideApplication.getActivityModule(this, getIntent(),
               aptoideApplication.getNotificationSyncScheduler(), aptoideApplication.getMarketName(),
               aptoideApplication.getAutoUpdateUrl(), (View) this,
-              aptoideApplication.getDefaultThemeName(), aptoideApplication.getDefaultStoreName(),
+              aptoideApplication.getDefaultThemeName(),
```
franciscoaleixo: Seems like this should be the getDefaultStoreName

PR #886, opened 2019-02-05
update android studio, dexcount, and bintray plugin versions

TacoTheDank:
- Android Studio 3.2.1 -> ~~3.3.0~~ 3.3.1
- Getkeepsafe's dexcount-gradle-plugin 0.6.1 -> 0.6.2 (latest version is 0.8.6)
- Jfrog's gradle-bintray-plugin 1.7.3 -> 1.8.4
- Added note to eventually delete Retrolambda
- Delete ```maven { url 'https://maven.google.com' }``` as it is replaced by just ```google()```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,30 +5,28 @@ buildscript {
     maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
     jcenter()
   }
+
   dependencies {
-    classpath 'com.android.tools.build:gradle:3.2.1'
-    classpath 'me.tatarka:gradle-retrolambda:3.7.0'
-    classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.6.1'
-    classpath 'io.realm:realm-gradle-plugin:3.5.0'
-    classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3'
+    classpath 'com.android.tools.build:gradle:3.3.1'
+    classpath 'me.tatarka:gradle-retrolambda:3.7.0' //TODO: Eventually delete this
+    classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.6.2' //0.8.6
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
```
jdandrade: We currently use a java codestyle file via Android Studio settings which is currently under aptoide-client-v8/codestyle/AptoideSquareAndroid_v3.xml
You can import this one for this project only and codestyle will be the same as we use today. 

In the future we might think about using the editorconfig file to enforce codestyle since it gives more flexibility and we can potentially use pre-commit git hooks to check codestyle.

For now I would just remove this file and use the AptoideSquareAndroid_v3.xml

TacoTheDank: The thing is, most people (like me) use a indent size of 4 as default. When I reformat code like this, it really dirties up the commits, and I'd hate for that to happen. Android Studio (for me) has never adopted this codestyle file for the particular project. Adding a .editorconfig means I won't have to worry about Android Studio automatically indenting everything to 4 spaces when I reformat (and then 20 minutes of individually reverting the changed indents), and you wouldn't have to worry about sifting through 1000 changed lines of reformatted indents looking for the 5 or so actual changes that were made.

Personally I'd prefer this project (and all projects) use 4 lines, as it's much easier to read, but that's up to the developers. I don't want to make that choice for you, because codestye is personal preference, so this file solves all that.

TacoTheDank: I don't think it would hurt to have both the xml codestyle file and the .editorconfig file.

jdandrade: Problem is that, from what I've experienced in your branch, the .editorconfig file is overriding the project's codestyle XML, which forces us to not be able to use the codestyle we created for us. 

The way I see this is that this is something that we should look into in a near future because it is, as you said, a more standardised way of having a codestyle across editors. But for now I will ask to remove the file and I can check codestyle in the reviews.

TacoTheDank: kk

PR #888, opened 2019-02-06
[ASV-1427] Created event to tell us whether apps in promotion need mi…

paquiquinho:
**What does this PR do?**

This PR aims to send a new event so we know, when an update is available in the promotions view, if the update needed would need to be migrated. The event has two fields: the package name of the app and whether it has the same signature as the app to download (this is how we check if play->catappult migration is needed, if the signatures are different)

**Database changed?**

No

- [ ] PromotionsAnalytics.java
- [ ] PromotionsPresenter.java

Since no app being sent to us that has a different signature we need to use charles to test this.
First, to see promotions, you need to put a breakpoint in the promotions request and replace the aptoide_md5 with "00000000000000000000000000000000" (the number of 0s is important). From there, download previous versions of a couple of the apps. On the response of the request where you changed the md5, change the signature of one of the apps that you downloaded. After seeing the promotions view in the scenario described you should see 2 events going out to our BI and 2 (equal) to facebook. The one you changed the signature should have the "signature" field filled with "different" while the other update app should have that field filled with "same".

  Tickets related to this pull-request: [ASV-1427](https://aptoide.atlassian.net/browse/ASV-1427)

**Code Review Checklist**

- [x] Architecture
- [x] Database changed?
- [x] Codestyle
- [x] Unit tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/promotions/PromotionsManager.java
+++ b/app/src/main/java/cm/aptoide/pt/promotions/PromotionsManager.java
@@ -10,6 +10,7 @@ import cm.aptoide.pt.download.AppContext;
 import cm.aptoide.pt.download.DownloadFactory;
 import cm.aptoide.pt.install.InstallAnalytics;
 import cm.aptoide.pt.install.InstallManager;
+import cm.aptoide.pt.install.InstalledRepository;
 import cm.aptoide.pt.notification.NotificationAnalytics;
 import java.util.List;
 import rx.Completable;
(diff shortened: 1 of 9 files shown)


app/src/main/java/cm/aptoide/pt/promotions/PromotionsManager.java
```diff
@@ -144,4 +147,15 @@ public String getWalletAddress() {
...
     return promotionsService.claimPromotion(walletAddress, packageName, captcha);
   }
+
+  public Observable<String> getPackageSignature(String packageName) {
+    return installedRepository.getInstalled(packageName)
+        .map(installed -> {
+          if (installed != null) {
+            return installed.getSignature();
+          } else {
+            return "";
```
jdandrade: This will always lead to a signature mismatch.

PR #905, opened 2019-02-18
[ASV-1350] Hide TC and PP from cobrand flavor.

tiagopedrinho:
**What does this PR do?**

   This PR aims to remove the TC and PP from the cobrand flavor. This was achieved by handling the show/hide on LoginSignupCredentials and PaymentLogin with two different presenters, one for each flavor. These presenters extend an abstract "base" presenter. For the settings part, simply used the manager created earlier to show/hide the preferences entries.

**Database changed?**

   No

- [ ] LoginSignupCredentialsFragment.java
- [ ] LoginSignupCredentialsPresenter.java
- [ ] LoginSignupCredentialsFlavorPresenter.java
- [ ] LoginSignupCredentialsPresenter.java
- [ ] LoginSignupCredentialsPresenter.java
- [ ] PaymentLoginFragment.java
- [ ] PaymentLoginPresenter.java
- [ ] PaymentLoginFlavorPresenter.java
- [ ] SettingsFragment.java

  Check both cobrand and vanilla versions, and see if the TC and PP are being shown/hidden in the correct flavors. Logins, settings.

  https://aptoide.atlassian.net/browse/ASV-1350

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle

### DIFF
--- a/app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
+++ b/app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
@@ -10,7 +10,7 @@ public class SettingsManager {
     return false;
   }
 
-  public boolean showGDRP() {
+  public boolean showGDPR() {
     return false;
   }
 }
--- a/app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
+++ b/app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
@@ -9,4 +9,8 @@ public class SettingsManager {
   public boolean showAdultContent() {
     return false;
   }
+
+  public boolean showGDRP() {
+    return false;
+  }
 }
--- a/app/src/cobrand/java/cm/aptoide/pt/billing/view/login/PaymentLoginFlavorPresenter.java
+++ b/app/src/cobrand/java/cm/aptoide/pt/billing/view/login/PaymentLoginFlavorPresenter.java
@@ -1,0 +1,142 @@
+package cm.aptoide.pt.billing.view.login;
+
+import cm.aptoide.accountmanager.AptoideAccountManager;
+import cm.aptoide.pt.account.AccountAnalytics;
+import cm.aptoide.pt.account.view.AccountNavigator;
+import cm.aptoide.pt.crashreports.CrashReport;
+import cm.aptoide.pt.orientation.ScreenOrientationManager;
+import cm.aptoide.pt.presenter.View;
+import cm.aptoide.pt.view.ThrowableToStringMapper;
+import java.util.Collection;
+import rx.Observable;
+import rx.Scheduler;
+
+public class PaymentLoginFlavorPresenter extends PaymentLoginPresenter {
+
+  private final PaymentLoginView view;
+  private final AccountAnalytics accountAnalytics;
+  private final AptoideAccountManager accountManager;
+  private final ThrowableToStringMapper errorMapper;
+  private final CrashReport crashReport;
+  private final ScreenOrientationManager orientationManager;
+  private final Scheduler viewScheduler;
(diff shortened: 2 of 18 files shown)


app/src/cobrand/java/cm/aptoide/pt/billing/view/login/PaymentLoginFlavorPresenter.java
```diff
@@ -0,0 +1,142 @@
+package cm.aptoide.pt.billing.view.login;
```
franciscoaleixo: Personally not a big fan of inheritance both here and on LoginSignupCredentialsPresenter for two reasons:

1) Conceptually there's no real reason to have inheritance in presenters other than repeated code because of different flavours. If we have X flavours, we will need X children. Even though unlikely to happen, it might present a scalability problem if we were to create other flavours.
2) It ironically leads to repeated code, check "handleAptoideSignUpEvent()" for example. The difference is two lines, the rest is repeated code. Again, if we need a new flavour we will have to repeat this because we likely will need to sign up.

Personally I prefer to isolate the differences as much as possible and avoid repeated code. How about having like a "PolicyManager" where it would hold whether it should show PP and TC and other related behaviour? Then you would only have one presenter that would react accordingly if we should show or not. This also gives flexibility if we want to change the visibility, for some reason, on the vanilla flavour.

app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
```diff
@@ -9,4 +9,8 @@ public SettingsManager() {
   public boolean showAdultContent() {
     return false;
   }
+
+  public boolean showGDRP() {
```
franciscoaleixo: GDPR*

app/src/cobrand/java/cm/aptoide/pt/billing/view/login/PaymentLoginFlavorPresenter.java
franciscoaleixo: I don't really like to use inheritance here for two reasons:

1. Conceptually there's no real reason for inheritance. The only advantage here is maybe to avoid duplicated code? Also, if we have X flavours, we will need X implementations of these presenters. Unlikely to be a problem but still a possible scalability issue.

2. It ironically leads to duplicated code. Check 'handleAptoideSignupEvent()', for example. The only difference between flavours is two lines and you need to implement the full method to even have signup. If we were to create a new flavour, we would need to add this method to have sign up. Not very good.

I do think composition is better here. Either have a 'PolicyManager' where it would hold whether we should show PP and TC or not, and other related behaviour. Consequently only have one presenter where it would act accordingly if we should or not show. This has the benefit of minimising repeated code, and also being able to hide the TC/PP on the vanilla flavour if there's any reason in the future.

app/src/cobrand/java/cm/aptoide/pt/SettingsManager.java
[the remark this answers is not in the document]
tiagopedrinho: Fixed.

tiagopedrinho: Fixed.

app/src/cobrand/java/cm/aptoide/pt/billing/view/login/PaymentLoginFlavorPresenter.java
tiagopedrinho: Removed unnecessary code, and extracted methods to prevent reuse of code. Also, opened a ticket for the future, for these classes to be refactored and remove the use of inheritance.

PR #925, opened 2019-02-28
[feature/ASV-1408] Play Protect Flag Analytics

franciscoaleixo:
**What does this PR do?**

   Implemented analytics on launch to check if the app is flagged by Play Protect using Google's SafetyNet API.

**Database changed?**

   No

- [ ] FirstLaunchAnalytics.java

  Check if the analytics are being sent for the 'Google_Play_Protect' event and verify different scenarios, e.g.: if play protect is not active, if aptoide is flagged and if aptoide is not flagged.

  [ASV-1408](https://aptoide.atlassian.net/browse/ASV-1408)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
+++ b/app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
@@ -64,19 +64,20 @@ public class FirstLaunchAnalytics {
 
   private final AnalyticsManager analyticsManager;
   private final AnalyticsLogger logger;
+  private final String packageName;
   private String utmSource = UNKNOWN;
   private String utmMedium = UNKNOWN;
   private String utmCampaign = UNKNOWN;
   private String utmContent = UNKNOWN;
   private String entryPoint = UNKNOWN;
-
   private SafetyNetClient safetyNetClient;
 
   public FirstLaunchAnalytics(AnalyticsManager analyticsManager, AnalyticsLogger logger,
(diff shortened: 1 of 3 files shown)


app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
```diff
@@ -74,6 +87,88 @@ public void sendFirstLaunchEvent(String utmSource, String utmMedium, String utmC
         FIRST_LAUNCH, AnalyticsManager.Action.OPEN, "");
   }
 
+  private void sendPlayProtectEvent() {
+    SafetyNet.getClient(context)
```
phillipsgoncalves: Like I said above, i would inject the safety net client on the class instead of getting it here using the context

app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
```diff
@@ -74,6 +87,88 @@ public void sendFirstLaunchEvent(String utmSource, String utmMedium, String utmC
...
+        .listHarmfulApps()
+        .addOnCompleteListener(task -> {
+          String isActive = "false";
```
phillipsgoncalves: To make this cleaner i would still use boolean objects and then i would convert to true or false where i am putting the value.
Example: isActive ? "true" : "false"

PR #929, opened 2019-03-04
[feature/ASV-1469] Search AB Test Parameter

franciscoaleixo:
**What does this PR do?**

   This PR tackles 3 different things: verifies a new parameter sent in the search request (items) and only sends action events if position<=items in list. It also adds two new analytic parameters for both impression and actions. Finally it handles situations where the search AB test cannot be retrieved (e.g. 404).

**Database changed?**

   No

  Check if analytic parameters are being sent correctly, check if actions are only recorded if the app is in position <= items (sent in the response) and check with Charles how the app behaves when the search_ab_test_mobile request cannot be retrieved (e.g. 404). Previously the search would not work when a 404 happened.

  [ASV-1469](https://aptoide.atlassian.net/browse/ASV-1469)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/search/analytics/SearchAnalytics.java
+++ b/app/src/main/java/cm/aptoide/pt/search/analytics/SearchAnalytics.java
@@ -15,6 +15,8 @@ public class SearchAnalytics {
   public static final String NO_RESULTS = "Search_No_Results";
   public static final String APP_CLICK = "Search_Results_App_View_Click";
   public static final String SEARCH_START = "Search_Start";
+  public static final String AB_SEARCH_ACTION = "AB_Searction_Action";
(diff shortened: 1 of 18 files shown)


app/src/main/java/cm/aptoide/pt/search/analytics/SearchAnalytics.java
```diff
@@ -15,6 +15,8 @@
   public static final String NO_RESULTS = "Search_No_Results";
   public static final String APP_CLICK = "Search_Results_App_View_Click";
   public static final String SEARCH_START = "Search_Start";
+  public static final String AB_SEARCH_ACTION = "AB_Searction_Action";
```
paquiquinho: AB_Searction_Action instead of Ab_Search_Action

PR #938, opened 2019-03-13
[ASV-1403] wallet ads offer

phillipsgoncalves:
**What does this PR do?**

   This PR aims at implementing the wallet ads offer card in home, implementing the new blacklist logic in the client and changing analytics events to manage the wallet ads offer.
   Regarding the wallet ads offer card in home, we were supposed to show a card in home telling the user that if he installs the wallet, he will stop seeing ads (mopub ads, the legacy ones should stay visible).
   Regarding the blacklist, a blacklist manager was created to start managing the blacklist clientside, since the blacklist services we have are being deprecated.
   Then, the download complete event was changed to have an extra attribute indicating if the user saw ads or not due to the wallet ads offer promotion. This event was only added to the users that were in group B of the ads ab-test (users that should see mopub ads). We also added a new event, with no parameters, in case the user does not see an interstitial ad due to the wallet promotion offer (just like the previous one, this event should only be sent in case the user is in the group b of the mopub ab test).

**Database changed?**

   No

- [ ] BundlesResponseMapper.java

There are 3 main things to test in this PR:
- Home wallet ads offer card: Test the interactions with the card. You should see the card in home in case the wallet is not installed and the card is not blacklisted. If one of these conditions is met, then the card should not be shown. Check that if you install the wallet, then you will stop seeing ads and that you will stop seeing the card showing in home. Please also confirm that all the events that are being sent for any card of the home are being sent.

- BlacklistManager in the client: You can use the wallet ads offer card to test this. Confirm that after 10 impressions the card disappears (is blacklisted). Test the "not now" button as well to confirm that the card becomes blacklisted after pressing the "not now" button.

- Check the download complete event. A new attribute was added to the download complete event. Note that this event was added everywhere where there are downloads, so please test it in Apps (update and update all), Promotions (download) and AppView download. Test this flow both ways, with the promotion "inactive" (wallet uninstalled so that you can see ads) and with the promotion active (so that you see ads) and check that the event is being correctly sent.
Another event that you need to test is the Ads_Wallet_Promotion event which is being sent everytime a user **DOES NOT SEE** an interstitial in AppView due to the ads wallet promotion.

Note: While developing the card in home I had to re-use some code from the AppCoins card. This means that i had to refactor a bit and extract some code from that card, so i would recommend taking a look at that card as well in order to guarantee that everything is ok with that card too.

  Tickets related to this pull-request: [ASV-1403](https://aptoide.atlassian.net/browse/ASV-1403)

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle


app/src/main/java/cm/aptoide/pt/home/BundlesResponseMapper.java
```diff
@@ -78,8 +85,14 @@ public BundlesResponseMapper(String marketName, InstallManager installManager,
               new Event().setName(Event.Name.getAds), widgetTag));
         } else if (type.equals(HomeBundle.BundleType.INFO_BUNDLE) || type.equals(
             HomeBundle.BundleType.EDITORIAL)) {
-          appBundles.add(new ActionBundle(title, type, event, widgetTag,
-              map((ActionItemResponse) viewObject)));
+          ActionItem actionItem = map((ActionItemResponse) viewObject);
+          appBundles.add(new ActionBundle(title, type, event, widgetTag, actionItem));
+        } else if (type.equals(HomeBundle.BundleType.WALLET_ADS_OFFER)) {
+          ActionItem actionItem = map((ActionItemResponse) viewObject);
+          if (walletAdsOfferCardManager.shouldShowWalletOfferCard(
```
franciscoaleixo: I will forever be against these identifier arguments (e.g. shouldShowWalletOfferCard(type+"_"+id). I know we're doing similarly in a lot of other cases so I'm not going to say for you to change it but the reason I don't like it is:
- For us to properly use 'shouldShowWalletOfferCard' and 'blackListManager.addImpression' we need to know the identifier structure, i.e. type + "_" + cardId, which makes using this harder. But if we pass the arguments and let those methods create the structure, we'll lose flexibility.

So I guess what I'm saying is that it's up to you :D

PR #939, opened 2019-03-13
[bugfix/ASV-1443] - Facebook OAuth request for cobrand flavour

franciscoaleixo:
**What does this PR do?**

   This PR changes the Facebook OAuth request for cobrand flavour, since before we were using the same request for cobrand/vanilla which doesn't work because we have a separate app id for cobrand versions.

**Database changed?**

   No

  Check if the facebook login works for both vanilla and cobrand flavours.

  [ASV-1433](https://aptoide.atlassian.net/browse/ASV-1433)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle


dataprovider/src/main/java/cm/aptoide/pt/dataprovider/ws/v3/OAuth2AuthenticationRequest.java
```diff
@@ -50,7 +51,9 @@ public static OAuth2AuthenticationRequest of(String username, String password, S
           body.put("oauthToken", password);
           break;
         case "FACEBOOK":
-          body.put("authMode", "facebook");
+          String authVal = "facebook";
+          if (cobrand) authVal = "facebook_cobrand";
```
phillipsgoncalves: I think you didn't need this. I would just put a provider for this string on the flavour application module and then on cobrand you would return "facebook_cobrand" and on vanilla you would return "facebook". Then, instead of injecting a boolean saying if it is cobrand you would inject the authValue that you need to use directly.

PR #941, opened 2019-03-14
[Feature/ASV-1370] Setting InstallerPackageName

jdandrade:
**What does this PR do?**

   Goal is to set the installer package name in the installed app so it can be checked with PackageManager.getInstallerPackageName() so the Cobrand product can make the APPC Attribution.
   This is only currently done in the vanilla internal/dev and cobrand flavours.
   Because of the new installation method that is used in these cases, there was a need to manage the installing state and to show it in the available views.
   To check the installer package name use an app named "Get Installer Package Name". You need to know the installed apps package name in order to check the installer package name.

**Database changed?**

   Yes

- [ ] DefaultInstaller.java
- [ ] AppInstaller.java

  Tickets related to this pull-request: [ASV-1370](https://aptoide.atlassian.net/browse/ASV-1370)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle


app/src/main/java/cm/aptoide/pt/ApplicationModule.java
```diff
@@ -315,12 +318,16 @@ public ApplicationModule(AptoideApplication application, String aptoideMd5sum) {
       @Named("secureShared") SharedPreferences secureSharedPreferences,
       DownloadsRepository downloadsRepository, InstalledRepository installedRepository,
       @Named("cachePath") String cachePath, @Named("apkPath") String apkPath,
-      @Named("obbPath") String obbPath, DownloadAnalytics downloadAnalytics) {
+      @Named("obbPath") String obbPath, AppInstaller appInstaller,
+      AppInstallerStatusReceiver appInstallerStatusReceiver,
+      InstallerPackageNameManager installerPackageNameManager) {
```
phillipsgoncalves: I don't know if this name is the best. I understand that it means that it is the installer where we will add the packagename on the installation but i don't know a better name.

jdandrade: I agree, can't find a better name than this one: PackageInstallerManager

packageinstaller/src/main/java/cm/aptoide/pt/packageinstaller/InstallResultReceiver.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP) @Override
+  public void onReceive(Context context, Intent intent) {
+    Bundle extras = intent.getExtras();
+    if (extras != null && "install_session_api_complete".equals(intent.getAction())) {
+      int status = extras.getInt(PackageInstaller.EXTRA_STATUS);
+      String message = extras.getString(PackageInstaller.EXTRA_STATUS_MESSAGE);
+      switch (status) {
+        case PackageInstaller.STATUS_PENDING_USER_ACTION:
+          // This app isn't privileged, so the user has to confirm the install.
```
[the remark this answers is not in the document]
jdandrade: this is a very valuable information, but since it is not exposed and the enum variable has documentation, I removed.

PR #952, opened 2019-03-22
[ASV-1403] Fix wallet ads offer events

phillipsgoncalves:
**What does this PR do?**
This PR aims at fixing wallet ads offer analytics events.
If the wallet ads offer is active then an attribute to the download complete event is added saying if the offer blocked the ads (true) or not (false).
Also, if the user does not see an interstitial ad due to the offer, an event without parameters is sent (Ads_Block_By_Offer).

**Database changed?**

   No

- [x] AppViewAnalytics.java
- [x] DownloadAnalytics.java

Do a download, check that the event is sent with the ads_block_by_offer false. Then install the wallet. Do another download, check that the same parameter is with true. Also, no interstitial ad should be shown there and the Ads_Block_By_Offer event should also be sent.

  Tickets related to this pull-request: [ASV-1403](https://aptoide.atlassian.net/browse/ASV-1403)

**Questions:**

**Code Review Checklist**

- [x] Architecture
- [x] Database changed?
- [x] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
+++ b/app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
@@ -53,7 +53,7 @@ public class DownloadAnalytics implements cm.aptoide.pt.downloadmanager.Analytic
   private static final String TELECO = "teleco";
   private static final String TRUSTED_BADGE = "Trusted Badge";
   private static final String URL = "url";
-  private static final String ADS_WALLET_PROMOTION = "ads_wallet_promotion";
+  private static final String ADS_BLOCK_BY_OFFER = "ads_block_by_offer";
   private final Map<String, DownloadEvent> cache;
   private final ConnectivityManager connectivityManager;
   private final TelephonyManager telephonyManager;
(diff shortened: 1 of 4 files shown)


app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
```diff
@@ -266,7 +266,7 @@ public void installClicked(String md5, String packageName, String trustedValue,
       downloadCompleteEvent(navigationTracker.getPreviousScreen(),
           navigationTracker.getCurrentScreen(), md5, packageName, trustedValue, action,
           previousContext,
-          offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.ADS_SHOW));
+          offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.ADS_HIDE));
```
jdandrade: does this make sense with the condition?` if (!offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.NO_ADS)) {
`

app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
```diff
@@ -282,7 +282,7 @@ public void installClicked(String md5, String packageName, AnalyticsManager.Acti
 
       downloadCompleteEvent(navigationTracker.getPreviousScreen(),
           navigationTracker.getCurrentScreen(), md5, packageName, null, action, previousContext,
-          offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.ADS_SHOW));
+          offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.ADS_HIDE));
```
jdandrade: same here:

`  if (!offerResponseStatus.equals(WalletAdsOfferManager.OfferResponseStatus.NO_ADS)) {`

app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
phillipsgoncalves: yes, because NO_ADS means that there are no ads (example ab test), so the parameter should not be added to the download complete event.

PR #964, opened 2019-04-09
[feature/ASV-1562] Ads User Property

franciscoaleixo:
**What does this PR do?**

   Adds user property for MoPub visibility status. It essentially tracks the APPC wallet installation when the app first starts and modifies the user property accordingly. Because of some dependency issues, async injection was implemented. Additionally, a new method in MoPubAdsManager was created to streamline its usage.

**Database changed?**

    No

  Check if the user property is being correctly set and also check if the modified events are still correct.

  [ASV-1562](https://aptoide.atlassian.net/browse/ASV-1562)

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle


app/src/main/java/cm/aptoide/pt/ApplicationComponent.java
```diff
@@ -26,4 +31,13 @@ ActivityComponent plus(ActivityModule activityModule,
...
   void inject(PullingContentService pullingContentService);
+
+  MoPubAdsManager getMopubAdsManager();
+
+  InstalledRepository getIdsRepository();
+
+  MoPubAnalytics getMoPubAnalytics();
+
+  CrashReport getCrashReport();
+
```
jdandrade: Why not inject this dependencies in the Application?

app/src/main/java/cm/aptoide/pt/AptoideApplication.java
```diff
@@ -344,6 +347,18 @@ public LeakTool getLeakTool() {
...
+        new FutureCallback<MoPubAdsService>() {
+          @Override public void onSuccess(@javax.annotation.Nullable MoPubAdsService service) {
+            if (service != null) {
+              service.observeWalletInstalation();
+            }
+          }
+
+          @Override public void onFailure(Throwable t) {
+          }
+        });
```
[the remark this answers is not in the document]
franciscoaleixo: I changed this completely so we don't compromise the architecture. The issue was that some of the MoPubAdsService dependencies depended on a Google Advertising ID fetch that needs to be retrieved in a background thread. Since regular injection runs on the main thread, this needed to be changed. The dependency tree of this service was also way too complex to initialize manually.

Even though I changed this so that the retrieval of the Google Advertising Id is now async inside the relevant dependency, it might be beneficial to explore asynchronous injection in the future (e.g to speed up launch time, which according to trace logs, the initial application injection setup takes about 400ms, not accounting for other dependency injections such as in activities, fragments...!).

Even though I used producers, there's more elegant ways to achieve async injection with RxJava explored here:
https://proandroiddev.com/dagger-2-part-three-new-possibilities-3daff12f7ebf

app/src/main/java/cm/aptoide/pt/ApplicationComponent.java
```diff
@@ -1,7 +1,12 @@
 package cm.aptoide.pt;
 
+import cm.aptoide.pt.ads.MoPubAdsManager;
+import cm.aptoide.pt.ads.MoPubAnalytics;
+import cm.aptoide.pt.crashreports.CrashReport;
 import cm.aptoide.pt.install.InstallService;
 import cm.aptoide.pt.install.InstalledIntentService;
+import cm.aptoide.pt.install.InstalledRepository;
+import cm.aptoide.pt.networking.IdsRepository;
```
jdandrade: remove unused imports

app/src/main/java/cm/aptoide/pt/ApplicationComponent.java
```diff
@@ -26,4 +31,5 @@ ActivityComponent plus(ActivityModule activityModule,
   void inject(InstalledIntentService installedIntentService);
 
   void inject(PullingContentService pullingContentService);
+
```
jdandrade: codestyle

app/src/main/java/cm/aptoide/pt/abtesting/ABTestService.java
```diff
@@ -20,35 +21,37 @@
...
-  public ABTestService(ServiceV7 service, String aptoideId) {
+  public ABTestService(ServiceV7 service, IdsRepository idsRepository) {
     this.service = service;
-    this.aptoideId = aptoideId;
+    this.idsRepository = idsRepository;
   }
 
   public Observable<ExperimentModel> getExperiment(String identifier) {
-    return service.getExperiment(identifier, aptoideId)
+    return getAptoideId().flatMap(aptoideId -> service.getExperiment(identifier, aptoideId)
```
jdandrade: we are not making sure this (getAptoideId()) is not running in the main thread, this might be a problem

app/src/main/java/cm/aptoide/pt/ads/MoPubAdsService.java
```diff
@@ -0,0 +1,41 @@
...
+import cm.aptoide.pt.crashreports.CrashReport;
+import cm.aptoide.pt.install.InstalledRepository;
+import rx.Scheduler;
+import rx.subscriptions.CompositeSubscription;
+
+public class MoPubAdsService {
+
+  private static final String WALLET_PACKAGE = "com.appcoins.wallet";
+
+  private final CompositeSubscription subscriptions;
```
jdandrade: we only have 1 subscription here and we are not unsubscribing it since It should probably live through the application scope.

app/src/main/java/cm/aptoide/pt/ads/MoPubAdsService.java
```diff
@@ -0,0 +1,41 @@
...
+        .distinctUntilChanged()
+        .flatMap(__ -> moPubAdsManager.getAdsVisibilityStatus()
+            .toObservable())
+        .doOnNext(moPubAnalytics::setAdsVisibilityUserProperty)
+        .takeWhile(offerResponseStatus -> offerResponseStatus
+            != WalletAdsOfferManager.OfferResponseStatus.NO_ADS)
+        .subscribe(created -> {
+        }, error -> crashReport.log(error)));
+  }
+}
```
jdandrade: this class name is strange since its purpose is to manage the Ads visibility user property. I would suggest something like AdsUserPropertyManager or something like this and the method name would be something like start().
observeWalletInstalation implies observation and it doesn't return an Observable nor an object, so start would be more appropriate I guess, if we can give a good name to the class.

app/src/main/java/cm/aptoide/pt/ads/MoPubAnalytics.java
```diff
@@ -12,4 +12,39 @@ public void setMoPubAbTestGroup(boolean isControlGroup) {
...
+  private AdsVisibility mapToAdsVisibility(WalletAdsOfferManager.OfferResponseStatus status) {
+    switch (status) {
+      case NO_ADS:
+        return AdsVisibility.CONTROL;
+      case ADS_HIDE:
+        return AdsVisibility.ADS_BLOCKED_BY_OFFER;
+      case ADS_SHOW:
+        return AdsVisibility.HAS_ADS;
+      default:
+        return AdsVisibility.CONTROL;
```
jdandrade: If this shouldn't happen I would throw an IllegalStateException or an error so we don't pollute analytics with data that could be not correct

PR #975, opened 2019-04-17
[AS-1612] Fix Load more on editorial view loading 2 of the same editorial content

JDSM01:
**What does this PR do?**

 Fixes the load more loading 2 of the same editorial content after a pull to refresh

**Database changed?**

No

-EditorialListAdapter

Check ticket

https://aptoide.atlassian.net/browse/ASV-1612

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/editorialList/EditorialListAdapter.java
+++ b/app/src/main/java/cm/aptoide/pt/editorialList/EditorialListAdapter.java
@@ -84,8 +84,7 @@ class EditorialListAdapter extends RecyclerView.Adapter<EditorialBundleViewHolde
   }
 
   public void update(List<CurationCard> curationCards) {
-    editorialListItems.clear();
-    editorialListItems.addAll(curationCards);
+    this.editorialListItems = curationCards;
     notifyDataSetChanged();
   }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/cm/aptoide/pt/editorialList/EditorialListAdapter.java
```diff
@@ -84,7 +84,8 @@ public synchronized int getLoadingPosition() {
   }
 
   public void update(List<CurationCard> curationCards) {
-    this.editorialListItems = curationCards;
+    editorialListItems.clear();
+    editorialListItems.addAll(curationCards);
```
franciscoaleixo: I don't think this is the correct fix, it's simply a workaround. The real reason why there's two of the same editorial content is because on EditorialListRepository.loadNewEditorialListViewModel, there's no cloning being done after caching the editorialListViewModel. This means that the adapter is still referencing the cache list view model. The cloning is important to dereference it from the cache.

Additionally, while loadMore should add items to the list, a pullToRefresh or a first load should not add to adapter, but rather replace it completely (i.e. call the adapter.update())

For reference this is how it's being done in HomeBundles, see: BundlesRepository and respective adapter on HomeFragment.

PR #979, opened 2019-04-18
[ASV-1502] Blacklist Appc Card

phillipsgoncalves:
**What does this PR do?**
This PR aims at implementing the blacklisting on the appc card. Since the mark as read api was deprected, we changed the blacklisting to the clientside. The code regarding the impressions of this card had to be adapted to this new clientside blacklist manager.

Also, i refactored the blacklist manager a bit in order to allow dynamic card id blacklisting.

**Database changed?**

No

- [x] BundlesResponseMapper.java
- [ ] Bar.kt

Since the APPC card is not live, you need to rewrite it on charles. Then scroll up and down to increase the impressions and check that after 10 impressions it was removed. Note: it is possible that it is cached so to check if it was correctly removed, you can just pull to refresh, which bypasses the cache.

  Tickets related to this pull-request: [ASV-1502](https://aptoide.atlassian.net/browse/ASV-1502)

**Questions:**

**Code Review Checklist**

- [x] Architecture
- [x] Database changed?
- [x] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/blacklist/BlacklistManager.java
+++ b/app/src/main/java/cm/aptoide/pt/blacklist/BlacklistManager.java
@@ -10,31 +10,32 @@ public class BlacklistManager {
     this.blacklistUnitMapper = blacklistUnitMapper;
   }
 
-  public boolean isBlacklisted(String blacklistId) {
-    return blacklister.isBlacklisted(blacklistUnitMapper.mapToBlacklistUnit(blacklistId));
+  public boolean isBlacklisted(String actionCardType, String id) {
+    return blacklister.isBlacklisted(
+        blacklistUnitMapper.mapActionCardToBlacklistUnit(actionCardType, id));
   }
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/cm/aptoide/pt/blacklist/BlacklistManager.java
```diff
@@ -10,31 +10,32 @@ public BlacklistManager(Blacklister blacklister, BlacklistUnitMapper blacklistUn
...
   }
 
-  public void blacklist(String blacklistId) {
-    blacklister.blacklist(blacklistUnitMapper.mapToBlacklistUnit(blacklistId));
+  public void blacklist(String actionCardType, String id) {
+    blacklister.blacklist(blacklistUnitMapper.mapActionCardToBlacklistUnit(actionCardType, id));
   }
 
-  public enum BlacklistUnit {
+  public enum BlacklistTypes {
```
jdandrade: Should follow class naming convention, non plural

PR #984, opened 2019-04-23
[ASV-1589] Fix MoPub lifecycle

phillipsgoncalves:
**What does this PR do?**

   This PR aims at passing the activity lifecycle to the mopub dependency. This is important to avoid having ads being emitted while the app is in background.

**Database changed?**

   No

- [ ] MainActivity.java

  Open the app and try to reproduce the old issue we had where we would emit interstitial ads after the app was already in background.

  Tickets related to this pull-request: [ASV-1589](https://aptoide.atlassian.net/browse/ASV-1589)

**Questions:**

**Code Review Checklist**

- [x] Architecture
- [x] Database changed?
- [x] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/view/MainActivity.java
+++ b/app/src/main/java/cm/aptoide/pt/view/MainActivity.java
@@ -12,7 +12,6 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.Resources;
 import android.os.Bundle;
-import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.design.internal.BottomNavigationItemView;
 import android.support.design.internal.BottomNavigationMenuView;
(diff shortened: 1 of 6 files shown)


app/src/main/java/cm/aptoide/pt/view/MainActivity.java
```diff
@@ -205,6 +224,30 @@ private void setupUpdatesNotification() {
...
+    MoPub.initializeSdk(this, sdkConfiguration, null);
+  }
+
+  @NonNull
+  private Map<String, String> getMediationNetworkConfiguration(String mediatedNetworkPlacementId) {
+    Map<String, String> mediationNetworkConfiguration = new HashMap<>();
+    mediationNetworkConfiguration.put("Placement_Id", mediatedNetworkPlacementId);
+    return mediationNetworkConfiguration;
+  }
+
```
jdandrade: Would keep this initialization in application since it was working and it is risky to delay this.

PR #987, opened 2019-04-29
[ASV-1593] Fixed native ads leaks in search

phillipsgoncalves:
**What does this PR do?**
This ticket aimed at investigating possible too many calls to the same ad unit. The problem could not be reproduced but there was a leak of native ads in search. This PR fixes that. Also removed unused code in home.

**Database changed?**

No

- [ ] SearchResultFragment.java

  Since the only changes were in search, open the app, go to the search view and check if there is anything wrong.

  Tickets related to this pull-request: [ASV-1593](https://aptoide.atlassian.net/browse/ASV-1593)

**Questions:**

**Code Review Checklist**

- [ ] Architecture
- [ ] Database changed?
- [ ] Codestyle

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/HomeFragment.java
+++ b/app/src/main/java/cm/aptoide/pt/home/HomeFragment.java
@@ -45,7 +45,6 @@ import rx.subjects.PublishSubject;
 public class HomeFragment extends NavigationTrackFragment implements HomeView {
 
   private static final String LIST_STATE_KEY = "cm.aptoide.pt.BottomHomeFragment.ListState";
-  private static final String MOPUB_CONSENT_DIALOG_KEY = "MoPub Consent Dialog";
 
   /**
    * The minimum number of items to have below your current scroll position before loading more.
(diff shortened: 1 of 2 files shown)


franciscoaleixo:
On similar apps, the mopub adapter also seems to leak. This is mostly because I also forgot to do it when I refactored similar apps, but maybe this would be a good time to also do this? I.e. in SimilarAppsViewHolder, the MoPubRecyclerAdapter is never destroyed.

phillipsgoncalves:
> On similar apps, the mopub adapter also seems to leak. This is mostly because I also forgot to do it when I refactored similar apps, but maybe this would be a good time to also do this? I.e. in SimilarAppsViewHolder, the MoPubRecyclerAdapter is never destroyed.

Since it is on an adapter, we don't really have a way to destroy it. I think it might not leak because the adapter will be set to null. The only problem is the fact that we will not call the destroy on the MoPubRecyclerAdapter (which would break the connection that this adapter has for incoming data). 

One solution would be to listen to the onViewDetached, but that means that we would have to load again on the onViewAttached. This means that we would be breaking and creating a new connection to MoPub a lot of times while scrolling, which would have performance impact.

Another possible solution would be to pass an instance of the view interface to the adapter and consequently to the ViewHolder. This means that we would have a subscription to be managed on the ViewHolder, which does not seem ideal and might be too much considering the problem that we have at the moment.

franciscoaleixo:
Even though this is not directly related to the ticket. Isn't the result lists recycler views leaking? I.e. followedStoresResultList, allStoresResultList, suggestionsResultList and trendingResultList. They all have their adapters set, but never unset.

According to [this resource](https://github.com/airbnb/epoxy/wiki/Avoiding-Memory-Leaks
), this would leak (even though it is related to Epoxy, it generally talks about leaks on recyclerview).

PR #1013, opened 2019-05-22
[ASV-1570] Fix progress obbs

phillipsgoncalves:
**What does this PR do?**

   This PR is due to the re-open of the correspondent ticket. In the last pull request, we updated the download object to have the size of the download so that we could divide the current progress by the total size of the app, instead of the number of files like we used to do.
This PR aims to fix a problem where in the Updates we were counting only with the apk size instead of the whole file size. This means that updates with obbs would have inconsistent progress.

**PS:** Database was changed, make sure everything is working correctly.

**Database changed?**

   Yes | 

- [x] Update.java

  Update an app with obbs. Example App vault (15 mb with obb)

  Tickets related to this pull-request: [ASV-1570](https://aptoide.atlassian.net/browse/ASV-1570)

**Questions:**

**Code Review Checklist**

- [x] Codestyle
- [x] Functional tests pass


app/src/main/java/cm/aptoide/pt/download/DownloadFactory.java
```diff
@@ -109,7 +109,7 @@ public Download create(Update update, boolean isAppcUpgrade) {
               update.getMainObbMd5(), update.getPatchObbPath(), update.getPatchObbMd5(),
               update.getUpdateVersionCode(), update.getUpdateVersionName(), update.getMainObbName(),
               update.getPatchObbName()));
-      download.setSize(Math.round(update.getFileSize()));
+      download.setSize(Math.round(update.getSize()));
```
jdandrade: size is long now, not needed

PR #1016, opened 2019-05-24
[feature/ASV-1536] AppView migration promotion

franciscoaleixo:
**What does this PR do?**

   This generalizes promotions a bit better and specifically implements migration promotion and analytics. A new db object was needed do distinguish migrated apps from non-migrated ones.

**Database changed?**

   No

- [ ] AppViewManager.java

  Check every main flow and see if it behaves as it should, some flows include:
  - Rewrite request for a package where two promotions are active but one of them is already claimed. In this case no promotion should appear.
  - Rewrite request for a package where two promotions are active and unclaimed. Further, the app should be ready to be migrated. In this case, the promotion should pertain to migration promotion. I think you'll have to check this in the debugger. Or maybe check claim requests
  - Rewrite request for a package where only migration promotion is active. In this case, promotion should only show to apps able to migrated.
  - Try to do a complete flow test as well (i.e. install wallet + migrate).
  - Test analytics again since they were refactored/implemented.
  - ... and any other case it may break :D 

  [ASV-1536](https://aptoide.atlassian.net/browse/ASV-1536)

NOTE: Tests still missing...
**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


dataprovider/src/main/java/cm/aptoide/pt/dataprovider/ws/v7/promotions/GetPromotionAppsResponse.java
```diff
@@ -12,7 +12,7 @@ public GetPromotionAppsResponse() {
   public static class PromotionAppModel {
```
phillipsgoncalves: codestyle

gradle.properties
```diff
@@ -113,7 +113,6 @@ MOPUB_NATIVE_PLACEMENT_ID_PROD=0
 MOPUB_NATIVE_SEARCH_PLACEMENT_ID_PROD=0
 #PROMOTION IDS
 HOME_PROMOTION_ID=BONUS_CARNIVAL_19
```
phillipsgoncalves: missing database version bump

app/src/main/java/cm/aptoide/pt/app/AppViewManager.java
```diff
@@ -533,50 +540,103 @@ private void sendAdsBlockByOfferEvent() {
...
-        return new WalletPromotionViewModel(false);
-      }
-    }
+  /**
+   * If the user clicks before cachedPromotionViewModel is set, this will return false, even though
+   * it might be true later. We could call loadPromotionViewModel(), but in this case it will do a
+   * lot of work twice. I think it's acceptable to assume it's false if the promotion wasn't shown
+   * yet.
+   */
+  public boolean hasClaimablePromotion(Promotion.ClaimAction action) {
```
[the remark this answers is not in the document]
franciscoaleixo: We've talked about this, but just to sum it up: this might take a long while because there's two different requests for promotions (the get promotion and the get wallet app), so this might add a significant delay to the AppView load.

As it's stated in the method doc, I don't think it's a big deal. If the user clicks on install without first seeing the promotion, should we even count it? As it stands, the user wouldn't know that there's a promotion so it's weird to count it as such.

phillipsgoncalves:
Problems found while testing:
-> AppView (non appc app) does not have the margin between the install button and the versions layout.
-> Downloaded wallet one of the times and there was no progress (don't think it is related to you, but did you see this bug happening a lot of times while developing).
-> Wrong message on the wallet card due to the setupInactiveWalletOffer method (where we are only checking open)

phillipsgoncalves:
Change rule of if you claim one promotion, you cant claim the other one.

PR #1017, opened 2019-05-27
[ASV-1663] Download state after 100%

phillipsgoncalves:
**What does this PR do?**

   Initially this PR aimed at fixing the download staying at 100% problem, but since we could not reproduce that issue with a specific behavior, we just added logs so that the next time it happens we can correctly trace the issue.
In this PR we also took the chance to look into other problems related with the downloads and tried to optimize some flows. The other fixes were:
 - Checking if file is already in the final path before trying to move it.
 - Optimized install chain to remove unnecessary database access and verifications.
 - Fixed download state parsing (initial state was not being correctly parsed).
 - Attempted at fixing a problem where we wouldn't be able to resume or cancel a download because the app downloader object was null (added logs since I couldn't reproduce it no more. I think the fix of the previous line could have been the solution)
 - Fixed the state after the installing state ( download app, at the end it will go through the installing state and pop the install dialog, cancel. When you come back press the install button again and there, the pause button would disappear).

As a result of this PR we also created the following ticket [ASV-1704](https://aptoide.atlassian.net/browse/ASV-1704)

**Database changed?**

   No

- [ ] AptoideDownloadManager.java
- [ ] InstallManager.java

  Since there are a lot of fixes on the downloads I would recommend the following:
 - Perform a download, pause and resume (to try to force the appdownloader problem).
 - Check if the installing status is being correctly handled in both APPC apps and non appc apps (also check the state after the install)
 - See if there are no problems popping the install dialog, besides the problem that we could not reproduce in a consistent way, where the download would be stuck at 100% (if you get this, please call me to check the logs).
 

  Tickets related to this pull-request: [ASV-1663](https://aptoide.atlassian.net/browse/ASV-1663)

**Questions:**

**Code Review Checklist**

- [x] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/download/DownloadInstallationProvider.java
```diff
@@ -28,6 +29,7 @@
  */
 public class DownloadInstallationProvider implements InstallationProvider {
 
+  private static final String TAG = "AptoideDownloadManager";
```
franciscoaleixo: The tag doesn't match the class name. It refers to another class. Mistake maybe?

app/src/main/java/cm/aptoide/pt/download/DownloadInstallationProvider.java
```diff
@@ -45,7 +47,10 @@ public DownloadInstallationProvider(AptoideDownloadManager downloadManager,
   }
 
   @Override public Observable<Installation> getInstallation(String md5) {
+    Logger.getInstance()
+        .d(TAG, "Getting the installation " + md5);
     return downloadManager.getDownload(md5)
+        .doOnError(throwable -> throwable.printStackTrace())
```
franciscoaleixo: Doesn't really matter but we normally put errors at the end of the chain.

app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -584,19 +595,33 @@ public boolean wasAppEverInstalled(String packageName) {
   }
 
   public void moveCompletedDownloadFiles(Download download) {
-    for (final FileToDownload fileToDownload : download.getFilesToDownload()) {
+    if (!haveFilesBeenMoved(download)) {
```
franciscoaleixo: Why do we have to loop two times? Why don't we verify in the same loop?

And what happens if one file has been moved and others didn't? With this method, it will still try to move that one file that was already moved.

app/src/main/java/cm/aptoide/pt/install/installer/DefaultInstaller.java
```diff
@@ -105,7 +98,10 @@ public PackageManager getPackageManager() {
           installation.setType(Installed.TYPE_UNKNOWN);
           moveInstallationFiles(installation);
         })
-        .flatMap(installation -> isInstalled(md5).first()
+        .flatMap(installation -> Observable.just(
+            isInstalled(installation.getPackageName(), installation.getVersionCode()))
+            .onErrorReturn(throwable -> false)
```
franciscoaleixo: Same thing as before about onErrorReturn

app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
```diff
@@ -1762,8 +1762,7 @@ private void setDownloadState(int progress, DownloadModel.DownloadState download
         pauseDownload.setVisibility(View.GONE);
         cancelDownload.setVisibility(View.GONE);
         resumeDownload.setVisibility(View.GONE);
-        downloadProgressValue.setVisibility(View.INVISIBLE);
```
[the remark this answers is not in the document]
phillipsgoncalves: João said that there is no problem showing it and it will be reviewed by Luis later

app/src/main/java/cm/aptoide/pt/install/installer/DefaultInstaller.java
phillipsgoncalves: This onErrorReturn should be here because it is related to the method above. If  the isInstalled returns an error we want it to pass false to the layers bellow.

PR #1020, opened 2019-05-30
[ASV-1654] Fix home navigation

phillipsgoncalves:
**What does this PR do?**

   This PR aims at fixing the navigation in home and keeping the state of all home fragments (games, apps and home). When you navigate to another view and comeback, the previous view should keep its state.

**Database changed?**

 No

- [ ] HomeContainerNavigator.java

  Open app, scroll down and open an appview. Press the back button and the home fragment should in the same position as before. Also same thing should happen on any of the "chips" fragments

  Tickets related to this pull-request: [ASV-1654](https://aptoide.atlassian.net/browse/ASV-1654)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/HomeContainerNavigator.java
+++ b/app/src/main/java/cm/aptoide/pt/home/HomeContainerNavigator.java
@@ -10,17 +10,20 @@ import cm.aptoide.pt.store.view.StoreTabGridRecyclerFragment;
 public class HomeContainerNavigator {
 
   private FragmentNavigator fragmentNavigator;
+  private String homeTag;
+  private String gamesTag;
+  private String appsTag;
 
   public HomeContainerNavigator(FragmentNavigator fragmentNavigator) {
     this.fragmentNavigator = fragmentNavigator;
   }
 
   public void loadMainHomeContent() {
-    Fragment fragment = fragmentNavigator.peekLast();
-    if (fragment != null && fragment instanceof HomeFragment) {
+    Fragment fragment = fragmentNavigator.getFragment(homeTag);
+    if (fragment != null) {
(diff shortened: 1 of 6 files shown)


franciscoaleixo:
There's two behaviour issues that I don't think are completely related to the ticket, so I don't think they should be done here, but it's good to note:

- Bottom navigation shouldn't clear the entire back stack. In fact, the fragments in BN shouldn't be cleared at all (except Search, I guess), only added ones (such as AppViewFragment). BN fragments should behave the same way as we're doing now with chips. Google Play has this behaviour and I think it's the way to go.
- HomeFragment always has this obnoxious loading indicator even when it clearly shouldn't be needed. MoreBundleFragment in chips behaves correctly, but not HomeFragment for some reason.

app/src/main/java/cm/aptoide/pt/home/HomeContainerNavigator.java
```diff
@@ -9,37 +9,59 @@
...
   }
 
   public void loadMainHomeContent() {
-    fragmentNavigator.navigateToWithoutBackSave(new HomeFragment(), true);
+    Fragment fragment = childFragmentNavigator.getFragment(homeTag);
+    if (fragment != null) {
+      childFragmentNavigator.navigateToWithoutBackSave(fragment, true);
+    } else {
+      homeTag = childFragmentNavigator.navigateTo(new HomeFragment(), true);
+    }
```
franciscoaleixo: I don't think it's really needed but eventually it might make sense to extract this navigation behaviour to FragmentNavigator, especially if we have a similar behaviour with BN in the future. But for now I think it's fine

PR #1027, opened 2019-06-12
[feature/ASV-1734] Dialog Base View for Wallet Install Dialog

franciscoaleixo:
**What does this PR do?**

   This PR creates a base activity for the wallet install dialog. The way it's setup, there's a main container in the activity layout used to inflate the actual content of the view (similar to MainActivity). 

This activity uses a specific custom dialog theme ('Aptoide.DialogTheme'). This theme defines several aspects of the dialog, namely: the minimum width in portrait/landscape mode (%) according to Zeplin, the dialog background with rounded edges etc.

Finally, this activity is launched as a new task, and 'taskAffinity', 'noHistory' and 'excludeFromRecents' are used in combination to properly use this activity independently from Aptoide task itself. DeepLinkIntentReceiver, which launches the dialog activity also needs a custom 'taskAffinity'.

Relevant analytics were also implemented.

**Database changed?**

   No

- [x] AndroidManifest.xml
- [x] styles.xml
- [x] cm.aptoide.pt.wallet package

  Call the relevant intent from anywhere (and with Aptoide both active and stopped), and check if the dialog behaves as expected (i.e. if it's independent from Aptoide and doesn't leave any trace in the phone's 'Recent'). Also because taskAffinity for DeepLinkIntentReceiver was changed, please check if no other activity launched from this one was broken (i.e. check other intents).

To test the deeplink use:

adb shell am start -a "android.intent.action.VIEW" "market://details?id=com.appcoins.wallet\&utm_source=myappcoins"

There’s a backslash before '&' to escape the character.

  [ASV-1734](https://aptoide.atlassian.net/browse/ASV-1734)

**Code Review Checklist**

- [x] Codestyle
- [x] Functional tests pass

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -131,6 +131,14 @@
         android:windowSoftInputMode="adjustPan"></activity>
 
     <activity
+        android:name=".wallet.WalletInstallActivity"
+        android:excludeFromRecents="true"
+        android:noHistory="true"
+        android:taskAffinity=".wallet.WalletInstallActivity"
+        android:theme="@style/Aptoide.DialogTheme"
+        android:windowSoftInputMode="adjustPan"/>
+
+    <activity
         android:name="com.ironsource.sdk.controller.ControllerActivity"
         android:configChanges="orientation|screenSize"
         android:hardwareAccelerated="true"/>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -262,7 +262,7 @@
     <activity
         android:name=".DeepLinkIntentReceiver"
         android:exported="true"
-        android:taskAffinity=""
+        android:taskAffinity=".DeepLinkIntentReceiver"
         >
       <intent-filter>
         <action android:name="android.intent.action.VIEW"/>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -262,6 +262,7 @@
     <activity
         android:name=".DeepLinkIntentReceiver"
         android:exported="true"
+        android:taskAffinity=""
         >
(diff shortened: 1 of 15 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -254,6 +262,7 @@
     <activity
         android:name=".DeepLinkIntentReceiver"
         android:exported="true"
+        android:taskAffinity=""
```
jdandrade: Why can't this be the default? (package-name)

jdandrade: Not sure if it is related to this but currently, if an Aptoide Activity is already in background - the deep link brings the Aptoide up with the dialog on top, instead of just the dialog on top of the current context

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallPresenter.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class WalletInstallPresenter(val view: WalletInstallView, val navigator: WalletInstallNavigator) :
+    Presenter {
+
+  override fun present() {
+    showDialog()
+  }
+
+  private fun showDialog() {
+    view.getLifecycleEvent()
```
jdandrade: can this be replaced by property access = view.lifecycleEvent()?

app/src/main/AndroidManifest.xml
franciscoaleixo: If you mean without taskAffinity, it needs it because otherwise the dialog will not be independent from the activity and it will always show behind the dialog. Setting the affinity to "" means that this has no affinity. I'm not sure as to why this is required though.

Alternatively, setting it to android:taskAffinity=".DeepLinkIntentReceiver" should also work fine

PR #1034, opened 2019-06-19
[ASV-1720] Changed home bundles view holders to not use the onBind po…

paquiquinho:
**What does this PR do?**

This PR aims to change the way we pass viewholder positions to home bundles. Instead of using the position provided when onBind is called, we now call getAdapterPosition inside the viewHolder

**Database changed?**

  No

- [ ] BundlesAdapter.java

The main implications for most of these positions regard analytics. Since these are currently impossible to check in charles, I used break points in every view holder onMoreClick method and checked if it matched. You can also add logs in the same place and check if the value is correct;

  Tickets related to this pull-request: [ASV-1720](https://aptoide.atlassian.net/browse/ASV-1720)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/AdsBundleViewHolder.java
+++ b/app/src/main/java/cm/aptoide/pt/home/AdsBundleViewHolder.java
@@ -52,7 +52,7 @@ class AdsBundleViewHolder extends AppBundleViewHolder {
     appsList.setNestedScrollingEnabled(false);
   }
 
-  @Override public void setBundle(HomeBundle homeBundle) {
+  @Override public void setBundle(HomeBundle homeBundle, int position) {
     if (!(homeBundle instanceof AdBundle)) {
       throw new IllegalStateException(this.getClass()
           .getName() + " is getting non AdBundle instance!");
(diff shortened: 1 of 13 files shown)


app/src/main/java/cm/aptoide/pt/home/AdsBundleViewHolder.java
```diff
@@ -52,27 +52,27 @@ public AdsBundleViewHolder(View view, PublishSubject<HomeEvent> uiEventsListener
...
+  @Override public void setBundle(HomeBundle homeBundle) {
     if (!(homeBundle instanceof AdBundle)) {
       throw new IllegalStateException(this.getClass()
           .getName() + " is getting non AdBundle instance!");
     }
     bundleTitle.setText(
         Translator.translate(homeBundle.getTitle(), itemView.getContext(), marketName));
 
-    appsInBundleAdapter.updateBundle(homeBundle, position);
+    appsInBundleAdapter.updateBundle(homeBundle, getAdapterPosition());
```
franciscoaleixo: getAdapterPosition() only really needs to be used for listeners, as inconsistencies only arise there. 

I'm not 100% sure but I think it's better to use the position from onBindViewHolder for this line in particular, because it might be NO_POSITION if it's doing layout calculations.

This is also why in the listeners, a NO_POSITION check is also recommended. 

Check this https://youtu.be/imsr8NrIAMs?t=2149

PR #1035, opened 2019-06-19
[ASV-1719] Fixed bug with loading more in search;

paquiquinho:
**What does this PR do?**

This PR aims to fix a bug where load more wouldn't show at the end of the search RV;

**Database changed?**

No

- [ ] SearchResultFragment.java

Check the ticket and make sure it's not happening anymore;

  Tickets related to this pull-request: [ASV-1719](https://aptoide.atlassian.net/browse/ASV-1719)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/search/view/SearchResultFragment.java
+++ b/app/src/main/java/cm/aptoide/pt/search/view/SearchResultFragment.java
@@ -51,6 +51,7 @@ import cm.aptoide.pt.store.StoreTheme;
 import cm.aptoide.pt.view.BackButtonFragment;
 import cm.aptoide.pt.view.ThemeUtils;
 import cm.aptoide.pt.view.custom.DividerItemDecoration;
+import com.jakewharton.rxbinding.support.v7.widget.RecyclerViewScrollEvent;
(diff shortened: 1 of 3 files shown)


app/src/main/java/cm/aptoide/pt/search/view/SearchResultFragment.java
```diff
@@ -536,21 +536,19 @@ private void forceSuggestions() {
     }
   }
 
-  private Observable<Void> recyclerViewReachedBottom(RecyclerView recyclerView) {
+  private Observable<Object> recyclerViewReachedBottom(RecyclerView recyclerView) {
```
franciscoaleixo: This Observable<Object> looks really weird and doesn't seem to have a purpose. Maybe keep the Void observable? I.e. map it to null.

franciscoaleixo:
The loading seems to have a white background which does not match the actual background color. I guess we've never noticed because it never showed. 

Maybe we should change this to transparent? I say this so that it is consistent with our other lists. For reference, this can be found in search_ad_loading_list_item.xml

PR #1047, opened 2019-06-27
[ASV-1738] Handle download wallet view

phillipsgoncalves:
**What does this PR do?**

   This PR aims at implementing the download of the wallet through helping AppCoins wallet installation flow. This flow includes:

- Handle of the cancel button.
- Handle close of the dialog after the app is installed
- Handle navigation to the installed wallet view after the installation is successful.
- Corrected check if app is already installed when the deeplink is launched.
- Using attribution method of install.

**Database changed?**

   No

- [ ] WalletInstallActivity.java

Send deeplink: 
 adb shell am start -a "android.intent.action.VIEW" -d "market://details?id=com.appcoins.wallet\&utm_source=appcoinssdk\&app_source=cm.aptoide.pt.dev"

and it will navigate to the dialog. After opening this dialog, if the wallet is installed it will show a success view and if it is not then it will show a loading and it will start the download. This download only has the cancel option as opposed to all the others where we also have the pause button.

  Tickets related to this pull-request: [ASV-1738](https://aptoide.atlassian.net/browse/ASV-1738)
[ASV-1737](https://aptoide.atlassian.net/browse/ASV-1737)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/wallet/WalletInstallPresenter.kt
```diff
@@ -28,23 +56,61 @@ class WalletInstallPresenter(val view: WalletInstallView,
...
+    return Observable.defer {
+      if (walletInstallManager.shouldShowRootInstallWarningPopup()) {
+        view.showRootInstallWarningPopup()
+            ?.doOnNext { answer -> walletInstallManager.allowRootInstall(answer) }
+      }
+      Observable.just(walletApp)
+    }.observeOn(viewScheduler)
+        .flatMap {
+          permissionManager.requestDownloadAccessValidatingMobileData(permissionService)
+              .flatMap { success ->
```
franciscoaleixo: 'success' not used, can be removed.

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallPresenter.kt
```diff
@@ -28,23 +56,61 @@ class WalletInstallPresenter(val view: WalletInstallView,
...
+                permissionManager.requestExternalStoragePermission(permissionService)
+              }
+              .observeOn(Schedulers.io())
+              .flatMapCompletable { void ->
```
franciscoaleixo: 'void' not used, can be removed.

app/src/main/res/values/strings_settings.xml
```diff
@@ -10,7 +10,7 @@
   <string name="setting_mobile">Download only over wifi</string>
   <string name="settings_beta_filter_setting_title">Beta versions</string>
   <string name="settings_beta_filter_summary">You\'ll also see beta apps.</string>
-  
+
```
franciscoaleixo: :dancer:

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallActivity.kt
```diff
@@ -41,11 +45,59 @@ class WalletInstallActivity : ActivityView(), WalletInstallView {
...
+    Logger.getInstance()
+        .d("WalletInstallActivity", "download state is " + downloadModel!!.downloadState)
+    if (downloadModel!!.isDownloading) {
+      setDownloadProgress(downloadModel)
+    } else {
+      wallet_install_download_view.visibility = View.GONE
+      progressView.visibility = View.GONE
+      walletInstallViewGroup.visibility = View.VISIBLE
+    }
+  }
```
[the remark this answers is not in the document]
phillipsgoncalves: didn't need the verification. Probably some changes made latter on in the code led to not being necessary to verify null

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallPresenter.kt
```diff
@@ -1,22 +1,50 @@
...
 import cm.aptoide.pt.presenter.Presenter
 import cm.aptoide.pt.presenter.View
-import cm.aptoide.pt.promotions.PromotionsManager
 import cm.aptoide.pt.promotions.WalletApp
+import rx.Completable
 import rx.Observable
 import rx.Scheduler
+import rx.schedulers.Schedulers
 
 class WalletInstallPresenter(val view: WalletInstallView,
```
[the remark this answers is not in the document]
phillipsgoncalves: Since we didn't discuss how codestyle should be implemented, I don't think it is worth it to be changing it everywhere, having the risk of having to change it back.

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallPresenter.kt
```diff
@@ -1,22 +1,50 @@
...
     loadWalletInstall()
     handleCloseButtonClick()
+    handleCancelDownloadButton()
+  }
+
+  private fun handleCancelDownloadButton() {
+    view.lifecycleEvent
+        .filter { lifecycleEvent -> View.LifecycleEvent.CREATE == lifecycleEvent }
+        .flatMap { view.cancelDownloadButtonClicked() }
+        .flatMap { walletInstallManager.getWallet() }.first()
```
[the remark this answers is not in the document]
phillipsgoncalves: yes, because it is an observable and in this case we don't want to keep observing the download. We only got the app here, just because (like we are doing in appview) we want to have the wallet object to be able to call the remove method.

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallActivity.kt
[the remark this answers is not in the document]
franciscoaleixo: It's true that the downloadModel here will not be null (because in the presenter/walletmanager it seems like it always creates a DownloadModel). But that's a problem probably in my part (since I did it), I should've probably flagged it as non-nullable.

However, it's a more general problem. Since it could be null (according to the WalletApp POJO), we should still verify it and not force it. If for some reason the downloadModel is allowed to be null in the future, this could crash. It's just safer and it's the recommended way.

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallManager.kt
```diff
@@ -14,4 +33,70 @@ class WalletInstallManager(val configuration: WalletInstallConfiguration,
...
+    return Observable.just(
+        downloadFactory.create(
+            downloadStateParser.parseDownloadAction(DownloadModel.Action.INSTALL),
+            walletApp.appName,
+            walletApp.packageName,
+            walletApp.md5sum, walletApp.icon, walletApp.versionName, walletApp.versionCode,
+            walletApp.path, walletApp.pathAlt, walletApp.obb,
+            false, walletApp.size))
+        .flatMapSingle { download ->
+          moPubAdsManager.adsVisibilityStatus.doOnSuccess { responseStatus ->
```
franciscoaleixo: Forgot to flag this in the review. I think we should change this to getAdsVisibilityStatus(), since it's not a class field of MoPubAdsManager. The IDE will complain a bit, but I think it's better.

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallView.kt
```diff
@@ -15,6 +16,14 @@ interface WalletInstallView : View {
 
   fun dismissDialog()
 
+  fun showRootInstallWarningPopup(): Observable<Boolean>?
```
franciscoaleixo: Remove '?'

PR #1052, opened 2019-07-04
[Feature/ASV-1746] Seasonal Promotions server side

jdandrade:
**What does this PR do?**

   Goal of this is to not have any reference to seasonal promotion ID's, titles and background - dependencies we had in order to create new promotions server side. Now we should be able to create seasonal promotions server side without changing anything in the app.
   To not reference any ID, we had to make a request to a new webservice that gives us promotion id's given a promotion type.
   Views affected are both Home and Promotion view.

**Database changed?**

   No

- [ ] PromotionsManager.java
- [ ] Home.java
- [ ] PromotionsFragment.java 

  Home promotion dialog and Promotions view. Charles rewrite included in comments -> [ASV-1746](https://aptoide.atlassian.net/browse/ASV-1746)

  Tickets related to this pull-request: [ASV-1746](https://aptoide.atlassian.net/browse/ASV-1746)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/promotions/PromotionsManager.java
```diff
@@ -62,10 +63,18 @@ public PromotionsManager(PromotionViewAppMapper promotionViewAppMapper,
...
-        .map(
-            appsList -> new PromotionsModel(appsList, getTotalAppc(appsList), isWalletInstalled()));
+  public Single<PromotionsModel> getPromotionsModel(String promotionsType) {
+    return promotionsService.getPromotions(promotionsType)
+        .flatMap(promotions -> {
+          if (promotions.isEmpty()) {
+            return Single.just(new ArrayList<PromotionApp>())
+                .map(__ -> PromotionsModel.ofError());
+          }
+          PromotionMeta meta = promotions.get(0);
```
phillipsgoncalves: Why do you get the promotion 0 here ? weren't we supposed to send a model with all the promotions ?

app/src/main/java/cm/aptoide/pt/promotions/PromotionsService.java
```diff
@@ -101,6 +105,32 @@ public String getWalletAddress() {
...
 
+  public Single<List<PromotionMeta>> getPromotions(String type) {
+    return GetPromotionsRequest.of(type, bodyInterceptorPoolV7, okHttpClient, converterFactory,
+        tokenInvalidator, sharedPreferences)
+        .observe()
+        .map(promotionsResult())
+        .toSingle();
+  }
+
+  @NonNull private Func1<GetPromotionsResponse, List<PromotionMeta>> promotionsResult() {
```
phillipsgoncalves: Why not creating a method and injecting the result inside of it like we usually do ?

app/src/main/java/cm/aptoide/pt/promotions/PromotionsManager.java
jdandrade: Here we are fetching the first promotion from that promotion type. If there are 2 seasonal promotions at the same time, we will show the most recent one.

app/src/main/java/cm/aptoide/pt/promotions/PromotionsService.java
jdandrade: In my opinion it makes the getPromotions method cleaner, but I agree that it should be the same pattern we use everywhere else.

PR #1053, opened 2019-07-05
[bugfix/ASV-1696] Click on Apps cards

franciscoaleixo:
**What does this PR do?**

   This PR fixes app cards clicks on Apps. Before, only updates not in progress actually did anything. Now clicking any state of an update card / active download card goes to the app view of the app.

Note that it was decided that analytics wouldn't change.

**Database changed?**

   No

  Test the click of every state of Update/Active Download card. In any state, it should go the app view. Installed card does not apply.

  [ASV-1696](https://aptoide.atlassian.net/browse/ASV-1696)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


franciscoaleixo:
> There's an issue with view holders on error state. If you click it, it will navigate to the appview as it should but once there the update button stops working. To reproduce this, go to apps -> turn off internet -> try to update an app, it should go to error -> turn internet back on -> navigate to appview -> after the error dialog, update button will not work;
> 
> Also worth noting is that going to the same appview from a different path will have a working button after the error dialog, so this should be connected to the navigation;

The reason this happens is because Apps is a mess. When a download is started, the download is retrieved (without appId), it is added to the adapter, and the old card is removed (which had a appId). Because the download has no appId, it interferes with the navigation which requires appId. This means that you couldn't update in appview any time you attempted to start the update on Apps (not just on error state).

To workaround this without doing major changes to Apps, I'm now simply forcing navigation ALWAYS by md5. 🤷‍♂

PR #1055, opened 2019-07-05
[ASV-1757] Helping wallet install metrics

phillipsgoncalves:
**What does this PR do?**

   This PR aims at implementing the metrics needed to the helping AppCoins wallet installation feature.
Also fixed the context used on the download complete event that is sent facebook. Previously we were using the previousContext instead of the currentContext meaning that we are going to send the previous view instead of the current view where the download is starting. 

Also fixed the Install and Download events sent to the BI by adding the appc flags, which were not added in ASV-1591.

**Database changed?**

No

- [ ] WalletInstallAnalytics.java

  Perform wallet download and installation and see the events being sent in both Charles (BI) and analytics manager logs.

  Tickets related to this pull-request: [ASV-1757](https://aptoide.atlassian.net/browse/ASV-1757)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/wallet/WalletInstallAnalytics.kt
```diff
@@ -39,9 +52,37 @@ class WalletInstallAnalytics(val downloadAnalytics: DownloadAnalytics,
...
+    if (shouldRegister) {
+      if (historyTracker == null) {
+        throw RuntimeException("If "
+            + this.javaClass
+            .simpleName
+            + " should be logged to screen history, it has to return a value on method NavigationTrackFragment#getHistoryTracker")
+      }
+      navigationTracker.registerScreen(historyTracker)
+    }
+    shouldRegister = false;
```
franciscoaleixo: ";"

app/src/main/java/cm/aptoide/pt/wallet/WalletInstallAnalytics.kt
```diff
@@ -39,9 +52,37 @@ class WalletInstallAnalytics(val downloadAnalytics: DownloadAnalytics,
...
+  fun sendClickOnInstallButtonEvent(packageName: String, applicationPublisher: String) {
+    val map = HashMap<String, Any>()
+    map[TYPE] = "Install"
+    map[APPLICATION_NAME] = packageName
+    map[APPLICATION_PUBLISHER] = applicationPublisher
+    analyticsManager.logEvent(map, CLICK_INSTALL, AnalyticsManager.Action.CLICK,
+        VIEW_CONTEXT)
+  }
+
+  fun setupHistoryTracker() {
```
franciscoaleixo: Why is this here? If we want to register the screen, shouldn't it be in the activity?

phillipsgoncalves: This was done by suggestion from João.
Ideally we would have a base activity that would do this management, just like we have the NavigationTrackerFragment, but João said it was not worth the effort, considering we only have one more activity which is the MainActivity. That is why every other navigation track management is done in the respective Fragments.
For this case we did it here in the wallet analytics because we needed to add a flag (shouldRegister) and we had to add the dependency of NavigationTracker. It didn't make sense to make the activity dirty because of this.

franciscoaleixo: 👍

PR #1056, opened 2019-07-05
[bugfix/ASV-1465] - Fixed appview action bar collapse bug 

franciscoaleixo:
**What does this PR do?**

   This PR fixes a bug where if an user would scroll in donations list in appview, the action bar would not collapse correctly.

**Database changed?**

   No

- [x] AppViewFragment.java

  Go to an app with appcoins and without screenshots (e.g. Red runner) and scroll down touching the Top donations list. Before, the action bar would not collapse correctly, and it would overlap the view. Now it should collapse correctly.

  [ASV-1465](https://aptoide.atlassian.net/browse/ASV-1465)

**Code Review Checklist**

- [x] Codestyle
- [x] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
+++ b/app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
@@ -378,11 +378,7 @@ public class AppViewFragment extends NavigationTrackFragment implements AppViewV
     donationsProgress = view.findViewById(R.id.donations_progress);
     donationsListLayout = view.findViewById(R.id.donations_list_layout);
     listDonateButton = view.findViewById(R.id.donate_button);
-    LinearLayoutManager linearLayoutManager = new LinearLayoutManager(getContext()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
```diff
@@ -380,7 +380,7 @@
     listDonateButton = view.findViewById(R.id.donate_button);
     LinearLayoutManager linearLayoutManager = new LinearLayoutManager(getContext()) {
       @Override public boolean canScrollVertically() {
-        return false;
+        return true;
```
jdandrade: This will be the same as the default, can we remove this implementation?
inside we have:
` @Override
    public boolean canScrollVertically() {
        return mOrientation == VERTICAL;
    }`

franciscoaleixo: 👍

PR #1087, opened 2019-07-27
[Feature/ASV-1825] splash screen

jdandrade:
**What does this PR do?**

   Shows up a splash screen while app is being launched. The main goal here is for the user to see something right away when he clicks the app launcher to not have the misconception that he failed to click it.
   Because we had to implement this splash screen in a different way, the Cobrand splash screen was adapted. Old splashscreen code was removed. Now Powered By Aptoide is not translated - this was decided.

**Database changed?**

   No

- [ ] AndroidManifest.xml

  Clear data > run vanilla
  run vanilla again
  Clear data > run cobrand
  run cobrand again

  Tickets related to this pull-request: [ASV-1825](https://aptoide.atlassian.net/browse/ASV-1825)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/cobrand/res/values/themes.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+  <style name="AppBaseTheme.Splash">
+    <item name="android:windowBackground">@drawable/splash_screen</item>
+  </style>
+</resources>
```
franciscoaleixo: Do we actually need this file? I thought that having just a themes.xml in main would be enough, and then splash_screen drawable would differ between flavours.

jdandrade: SVG's doesn't seem to be supported as a bitmap source... I was getting errors because of it, hence why I used the pngs.

I think you are right about this cobrand file, I think that was some experiementation I was doing with different drawables and then I forgot to merge this file back into main, will test it now.

PR #1108, opened 2019-08-12
[ASV-1807] load more home

phillipsgoncalves:
**What does this PR do?**

   This PR aims at implementing the new error for home load more.

**Database changed?**

 No

- [ ] HomePresenter.java

Open Home, turn off wifi, go to load more section. See the error. Turn internet on again, press retry and it should load the next items from home.

  Tickets related to this pull-request: [ASV-1807](https://aptoide.atlassian.net/browse/ASV-1807)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/home/BundlesAdapter.java
```diff
@@ -198,4 +207,14 @@ public synchronized void updateEditorials() {
       }
     }
   }
+
+  public void showLoadMoreError() {
+    bundles.add(errorBundle);
+    notifyItemInserted(bundles.size());
```
franciscoaleixo: Should be bundles.size() - 1, although it's interesting that it works.

app/src/main/java/cm/aptoide/pt/home/ErrorHomeBundle.java
```diff
@@ -0,0 +1,7 @@
+package cm.aptoide.pt.home;
+
+public class ErrorHomeBundle extends DummyBundle {
```
franciscoaleixo: This wasn't you and you don't have to do anything about this... But it just seems unnecessary to have a DummyBundle just to avoid implementing null 🤣

app/src/main/java/cm/aptoide/pt/home/BundlesAdapter.java
franciscoaleixo: Also, even though this proably won't ever be a problem, it would be safer to do:

remove(bundles.indexOf(errorBundle))
add(bundles.indexOf(errorBundle))

Because this way you guarantee that you're notifying the correct item and you don't assume that it's at the end.

Remove method already exists within the class, although the add would need to be added for individual items.

app/src/main/java/cm/aptoide/pt/home/LoadingMoreErrorViewHolder.java
```diff
@@ -0,0 +1,23 @@
+package cm.aptoide.pt.home;
+
+import android.view.View;
+import android.widget.Button;
+import cm.aptoide.pt.R;
+import rx.subjects.PublishSubject;
+
+class LoadingMoreErrorViewHolder extends AppBundleViewHolder {
```
franciscoaleixo: Not in Kotlin... 🔔 Shame 🔔

PR #1119, opened 2019-08-25
[Feature/ASV-1905] Support Splits 

jdandrade:
**What does this PR do?**

   Shows Apps with splits in Search results, App view, Apps view, every list of apps (bundles, recommended).
   Potentially shows Apps with splits in Editorial content, Promotions (these need to be supported in the API side first).
   Adapted DownloadFactory to create Download's with base.apk+splits. This was where most of the work was done in order to allow splits download.
   Adapted AptoideDownloadManager to support split downloads (small changes to some mappers).
   Adapted InstallManager to support split installations and redirected the installs to the AppInstaller, which uses PackageInstaller to install apps with splits.
   Fixed an issue with moveCompletedDownloadFiles where it wasn't correctly moving apks and obbs after the first app was downloaded and correctly moved there.
   Database objects like Download / FileToDownload and Updates had to be changed to allow splits metadata.

   File system logic is now the following:
- Everything being downloaded is temporarly under .aptoide/ dir (this remains unchanged)
- Splits have the following name pattern: split-md5+split-name+.apk
- After App with splits is downloaded, splits are moved under: .aptoide/<app_name>-splits/

An app with splits is only valid if it respects the 
`"required_split_types": [
        "ABI",
        "DENSITY"
      ]`
Json in all aab responses.
The app is valid if ws response contains at least a split for each required split type. In this example the splits array would have to have at least an ABI and a DENSITY split to be considered valid. At the moment if it is not valid (it shouldn't happen since it should be correctly filtered in API) the download does not start.

In the realm migration I had to use a hack to have a RealmList<String> for the required splits. below realm 4.0.0 you cannot have a RealmList of primitive types, so I had to use an object (RealmString.java) with only a primitive type in it: String. The class already existed in the project and it was used somewhere in the class, that is why I check in the migration if the schema for that class already exists... Hopefully, this will all go away when we ditch realm.

  Installing state in app view was centered and is now larger, so it doesn't show that weird off-centered indeterminate state while installing. Also no progress is shown while installing.
  
  Some of the changes were Codestyle changes... I hope it's not a linux vs macos codestyle problem.

**Database changed?**

   Yes

- [ ] v7.java (this will be all the "enable" aab's in requests changes
- [ ] DownloadFactory.java (here you will have almost everything related to the download of splits logic + naming and file paths)
- [ ] All app network objects
- [ ] All view models and models that have app information
- [ ] DefaultInstaller.java
- [ ] RealmToRealmDatabaseMigration.java

  There is a store with an app with splits. The store was created by FReis and it is called "split-store". Currently it has 1 app: linkedin (com.linkedin.android). This app is version 4.1.335.1 and its only uploaded in this store. There is no superior version than this in Aptoide.
This version currently only has 2 splits added (an armv8a abi split and a xxhdpi density split), which means that you will **NOT** see it if you don't have a compatible phone (needs to be compatible with armv8a, usually an armv7a phone will be compatible, but a x86 [emulators usually] won't). 
Test compatible device > app shows up in splits-store.
Test uncompatible device > app does not show up (# of applications in store still is 1... which is weird for the user and probably a bug in the api).
This linkedin app should show up in Search results, app view, apps (if you download an earlier version and follow split-store).

After everything about showing up aab's is tested, test:
- [ ] Download com.linkedin.android with splits without linkedin installed. (check .aptoide in file system to see if files are being handled correctly)
- [ ] Update to com.linked.android with splits in appview and apps (install an earlier linked in version first without splits)
- [ ] Test Downgrade, from splits to older version.

  Tickets related to this pull-request: [ASV-1905](https://aptoide.atlassian.net/browse/ASV-1905)
  Documentation related to this pull-request: [App Bundles API](https://aptoide.atlassian.net/wiki/spaces/AN/pages/1358725187/App+Bundle+api)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/app/AppViewManager.java
```diff
@@ -262,7 +263,7 @@ public Completable downloadApp(WalletApp walletApp) {
             .getAction()), walletApp.getAppName(), walletApp.getPackageName(),
         walletApp.getMd5sum(), walletApp.getIcon(), walletApp.getVersionName(),
         walletApp.getVersionCode(), walletApp.getPath(), walletApp.getPathAlt(), walletApp.getObb(),
-        false, walletApp.getSize()))
+        false, walletApp.getSize(), null, null))
```
phillipsgoncalves: Minor,but why adding nulls here? Shouldn't we use it the same way as we are using for the other app models? That way, if appcoins wallet one day becomes an appbundle we won't have to change anything.

app/src/main/java/cm/aptoide/pt/editorial/EditorialManager.java
```diff
@@ -72,7 +72,7 @@ public Completable downloadApp(EditorialDownloadEvent editorialDownloadEvent) {
         editorialDownloadEvent.getMd5(), editorialDownloadEvent.getIcon(),
         editorialDownloadEvent.getVerName(), editorialDownloadEvent.getVerCode(),
         editorialDownloadEvent.getPath(), editorialDownloadEvent.getPathAlt(),
-        editorialDownloadEvent.getObb(), false, editorialDownloadEvent.getSize()))
+        editorialDownloadEvent.getObb(), false, editorialDownloadEvent.getSize(), null, null))
```
phillipsgoncalves: This is assuming that there will be no appbundles in editorial, right?

app/src/main/java/cm/aptoide/pt/app/AppViewManager.java
jdandrade: already discussed, done.

app/src/main/java/cm/aptoide/pt/editorial/EditorialManager.java
jdandrade: you are assuming wrong! it is there already

app/src/main/java/cm/aptoide/pt/editorial/EditorialRequest.java
```diff
@@ -46,6 +46,6 @@ public static EditorialRequest of(String cardId, BodyInterceptor<BaseBody> bodyI
 
   @Override protected Observable<EditorialCard> loadDataFromNetwork(Interfaces interfaces,
       boolean bypassCache) {
-    return interfaces.getEditorial(cardId, body);
+    return interfaces.getEditorial(cardId, body, true);
```
phillipsgoncalves: we are passing true in here, so we will be receiving app bundles

downloadmanager/src/main/java/cm/aptoide/pt/downloadmanager/DownloadAppFile.java
```diff
@@ -46,15 +46,16 @@ public String getPackageName() {
   }
 
   public int getFileType() {
-    return fileType.getType();
+    return fileType.
```
phillipsgoncalves: I don't know if these codestyle changes are from the different os but if I put everything the same way it was before (getType on the same line) and apply the codestyle nothing changes

aptoide-database/src/main/java/cm/aptoide/pt/database/realm/FileToDownload.java
```diff
@@ -199,7 +199,8 @@ public void setMd5(String md5) {
     this.md5 = md5;
   }
 
-  @IntDef({ APK, OBB, GENERIC }) @Retention(RetentionPolicy.SOURCE) public @interface FileType {
+  @IntDef({ APK, OBB, GENERIC, SPLIT }) @Retention(RetentionPolicy.SOURCE)
+  public @interface FileType {
```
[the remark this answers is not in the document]
jdandrade: It was already public, check above. codestyle problems.

aptoide-database/src/main/java/cm/aptoide/pt/database/realm/Split.java
```diff
@@ -0,0 +1,63 @@
+package cm.aptoide.pt.database.realm;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+
+public class Split extends RealmObject {
+  @PrimaryKey private String md5;
+  private String link;
```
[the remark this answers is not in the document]
jdandrade: done

dataprovider/src/main/java/cm/aptoide/pt/dataprovider/ws/v7/promotions/GetPromotionAppsRequest.java
```diff
@@ -30,7 +30,7 @@ public static GetPromotionAppsRequest of(String promotionId,
   @Override
   protected Observable<GetPromotionAppsResponse> loadDataFromNetwork(V7.Interfaces interfaces,
       boolean bypassCache) {
-    return interfaces.getPromotionApps(30, body, bypassCache);
+    return interfaces.getPromotionApps(30, true, body, bypassCache);
```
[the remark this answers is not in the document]
jdandrade: done

downloadmanager/src/main/java/cm/aptoide/pt/downloadmanager/DownloadAppFile.java
jdandrade: yea, probably a misclick - done

app/src/main/java/cm/aptoide/pt/install/installer/DefaultInstaller.java
```diff
@@ -301,7 +301,8 @@ private Installation updateInstallation(Installation installation, int type, int
                       .d("Installer", "status: " + installStatus.getStatus()
                           .name() + " " + installation.getPackageName());
                   updateInstallation(installation,
-                      shouldSetPackageInstaller ? Installed.TYPE_SET_PACKAGE_NAME_INSTALLER
+                      shouldSetPackageInstaller || !map(installation).getSplitApks()
```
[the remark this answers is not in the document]
jdandrade: done

app/src/main/java/cm/aptoide/pt/view/app/AppService.java
```diff
@@ -249,13 +251,29 @@ public AppService(StoreCredentialsProvider storeCredentialsProvider,
               app.hasBilling(), app.hasAdvertising(), app.getBdsFlags(), app.getAge()
               .getRating() == MATURE_APP_RATING, app.getFile()
               .getSignature()
-              .getSha1());
+              .getSha1(), app.hasSplits() ? map(app.getAab()
+              .getSplits()) : Collections.emptyList(), app.hasSplits() ? app.getAab()
```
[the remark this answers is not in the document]
jdandrade: In my opinion Aab pojo should reflect exactly what comes from the web, then we wrap it with other layers with that logic. I don't think it is worth the effort here. to make all the changes because of these.

phillipsgoncalves:
From the tests I did:
- the downgrade is not working correctly (it is only uninstalling the old version and not installing the new one afterwards) -> don't know if this is from dev or not.
- data base migration problems

jdandrade:
Fixed migration, Downgrade is working in the tests I've made. Installing state is not showing up on the Downgrade flow -> need to open a issue.

PR #1120, opened 2019-08-26
[feature/ASV-1859] Fixed tag for actions in bundles

phillipsgoncalves:
**What does this PR do?**
This PR aims at fixing the problems related with the tag of the "more section" bundles, which was being sent wrong in the events
   

**Database changed?**

   No

- [ ] AppBundle.java

Open appview from the editors choice and then open the appview through the more section. Check that the tags are correct.

  Tickets related to this pull-request: [ASV-1859](https://aptoide.atlassian.net/browse/ASV-1859)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/BundlesResponseMapper.java
+++ b/app/src/main/java/cm/aptoide/pt/home/BundlesResponseMapper.java
@@ -57,6 +57,7 @@ public class BundlesResponseMapper {
         Event event = getEvent(widget);
 
         String widgetTag = widget.getTag();
+        String widgetActionTag = getWidgetActionTag(widget);
 
         Object viewObject = widget.getViewObject();
         String title = widget.getTitle();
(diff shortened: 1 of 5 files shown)


app/src/main/java/cm/aptoide/pt/home/BundlesResponseMapper.java
```diff
@@ -66,14 +67,18 @@ public BundlesResponseMapper(InstallManager installManager,
         if (type.equals(HomeBundle.BundleType.APPS)
             || type.equals(HomeBundle.BundleType.EDITORS)
             || type.equals(HomeBundle.BundleType.TOP)) {
+
+          String widgetActionTag = getWidgetActionTag(widget);
           appBundles.add(new AppBundle(title, map(((ListApps) viewObject).getDataList()
-              .getList(), type, widgetTag), type, event, widgetTag));
+              .getList(), type, widgetTag), type, event, widgetTag, widgetActionTag));
         } else if (type.equals(HomeBundle.BundleType.APPCOINS_ADS)) {
+          String widgetActionTag = getWidgetActionTag(widget);
```
franciscoaleixo: Looks like you can just use one `String widgetActionTag = getWidgetActionTag(widget);` before all the if's instead of two.

app/src/main/java/cm/aptoide/pt/home/HomeNavigator.java
```diff
@@ -65,10 +65,15 @@ public void navigateWithDownloadUrlAndReward(long appId, String packageName, Str
...
+        .getTag();
+    if (click.getBundle() instanceof AppBundle) {
+      tag = ((AppBundle) click.getBundle()).getActionTag();
+    }
+
     fragmentNavigator.navigateTo(StoreTabGridRecyclerFragment.newInstance(click.getBundle()
         .getEvent(), click.getType(), click.getBundle()
-        .getTitle(), "default", click.getBundle()
-        .getTag(), StoreContext.home), true);
+        .getTitle(), "default", tag, StoreContext.home), true);
```
franciscoaleixo: Looks like we should pass tag and tag action since it's two different things... Because if not, what's the use of having the tag and not tag action in other cases other than AppBundle?

PR #1150, opened 2019-10-17
[ASV-1901] load more error bundle now working as intended;

paquiquinho:
**What does this PR do?**

This PR fixes an issue where load more wouldn't correctly show in home;

**Database changed?**

No

- [ ] RemoteBundleDataSource.java

Make sure the bug described in the ticket isn't happening anymore and that no side effects appear from the fix;

  Tickets related to this pull-request: [ASV-1901](https://aptoide.atlassian.net/browse/ASV-1901)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
+++ b/app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
@@ -535,6 +535,7 @@ public class HomePresenter implements Presenter {
   }
 
   private void handleLoadMoreError() {
+    home.setLoadMoreError();
     view.showLoadMoreError();
   }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -535,6 +535,7 @@ private void handleEditorialCardClick() {
   }
 
   private void handleLoadMoreError() {
+    home.setLoadMoreError();
```
phillipsgoncalves: I think this should be abstracted from the presenter. I would do this inside home itself, without having to come to the presenter.
Inside the loadNextHomeBundles i would check for the existence of an error and i would set it there.

PR #1160, opened 2019-10-28
[ASV-2040] AppNext for exclusive

phillipsgoncalves:
**What does this PR do?**

 This PR aims at implementing ads from AppNext to mature appviews.
To integrate AppNext i also had to increase play services location version to 17.0.0.

**Database changed?**

   No

- [ ] AppnextBaseConfigurationCustomEvent.java

Run the app check if you see ads on mature apps. If not this could be because of no fill. Check the logs to see if there is any error.

  Tickets related to this pull-request: [ASV-2040](https://aptoide.atlassian.net/browse/ASV-2040)

**Questions:**

Yes, new placements for the new ad units. 

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -199,21 +199,21 @@ private void sendAppViewLoadAnalytics(AppViewModel appViewModel) {
...
   }
 
-  private Observable<Boolean> loadInterstitialAds() {
+  private Observable<Boolean> loadInterstitialAds(AppModel appModel) {
     return appViewManager.shouldLoadInterstitialAd()
         .observeOn(viewScheduler)
         .flatMap(shouldLoad -> {
           if (shouldLoad) {
-            view.initInterstitialAd();
+            view.initInterstitialAd(appModel.isMature());
```
jdandrade: this method only depends on the mature flag, we can remove the heavier dependency to not couple this method to the AppModel

app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -222,12 +222,11 @@ private void sendAppViewLoadAnalytics(AppViewModel appViewModel) {
...
   }
 
-  private Observable<Boolean> loadBannerAds() {
+  private Observable<Boolean> loadBannerAds(AppModel appModel) {
     return appViewManager.shouldLoadBannerAd()
-        .observeOn(viewScheduler)
         .doOnSuccess(shouldLoadBanner -> {
           if (shouldLoadBanner) {
-            view.showBannerAd();
+            view.showBannerAd(appModel.isMature());
```
jdandrade: this method only depends on the mature flag, we can remove the heavier dependency to not couple this method to the AppModel

app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -897,6 +896,7 @@ private void handleClickOnRetry() {
             .getKeywords())
         .flatMap(similarAppsViewModel -> appViewManager.shouldLoadNativeAds()
             .doOnSuccess(similarAppsViewModel::setShouldLoadNativeAds)
+            .doOnSuccess(__ -> similarAppsViewModel.setFromMatureApp(appViewModel.isMature()))
```
jdandrade: Why are we updating this flag after object is instantiated when we should inject this info?

phillipsgoncalves: Stupidly I did not inject it because this model was also used for appcoins and there are no ads in the appcoins similar apps bundle, so at the time I thought I should not inject it as it would not depend on it.
Refactored it and also changed the should load native ads flag.

PR #1192, opened 2019-12-06
[feature/ASV-2119] ABTest apkfy

franciscoaleixo:
**What does this PR do?**

   This PR creates an AB Test where we have two groups: one where apkfy users will not see an interstitial on the first install, and one as normal (they will see an ad regardless). It also sends an event to Rakam and sets super properties to measure retention.

**Database changed?**

   No

  First, you have to build the code and insert apkfy related files to test it. So either test in dev version with v1 signing (+apkfy files) or test with release (+apkfy files), if you can see release requests in Charles. Next, you have to test both groups ('control' and 'no_ads_on_first_session') and check for 'control' that interstitial appears on the first install, and that it doesn't for 'no_ads_on_first_session'. Independently of this, check if an event was sent to Rakam analytics with the result of the AB Test. Also check if the current AB Test group is also set on Rakam super properties (and if you force stop aptoide, they should still be there).

  For testing both groups, you can use the rewrite on the JIRA ticket below.

  [ASV-2119](https://aptoide.atlassian.net/browse/ASV-2119)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -1073,7 +1075,7 @@ private void handleInstallButtonClick() {
                   completable = appViewManager.getAppModel()
                       .flatMapCompletable(appModel -> appViewManager.getAdsVisibilityStatus()
                           .flatMapCompletable(
-                              status -> downloadApp(action, appModel, status).observeOn(
+                              status -> downloadApp(action, appModel, status, false).observeOn(
```
phillipsgoncalves: Sending false here, while few lines bellow, for the click on install button it is checking the value.

phillipsgoncalves: Btw, resume download method , downgrade, etc should also send the correct apkfy value instead of false

app/src/main/java/cm/aptoide/pt/download/DownloadAnalytics.java
```diff
@@ -179,7 +180,7 @@ public void downloadStartEvent(Download download, AnalyticsManager.Action action
   }
 
   public void downloadStartEvent(Download download, int campaignId, String abTestGroup,
-      AppContext context, AnalyticsManager.Action action, boolean isMigration) {
+      AppContext context, AnalyticsManager.Action action, boolean isMigration, boolean isApkfy) {
```
phillipsgoncalves: unused variable

PR #1207, opened 2019-12-23
[ASV-2214] Rakam user properties

phillipsgoncalves:
**What does this PR do?**

   Sending all rakam user properties at once as rakam is paid by events, so if we sent all properties in different "events" we would be spending more.

**Database changed?**

No

- [ ] FirstLaunchAnalytics.java

  Run the app and check in Charles that the events are only sent once. 

  Tickets related to this pull-request: [ASV-2214](https://aptoide.atlassian.net/browse/ASV-2214)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/GmsStatusValueProvider.java
+++ b/app/src/main/java/cm/aptoide/pt/GmsStatusValueProvider.java
@@ -5,7 +5,7 @@ import cm.aptoide.pt.dataprovider.ads.AdNetworkUtils;
 
 public class GmsStatusValueProvider {
 
-  private static final String HAS_HGMS = "Has GMS";
+  private static final String HAS_GMS = "Has GMS";
   private static final String NO_GMS = "No GMS";
   private Context context;
   private String gmsValue = "";
@@ -16,7 +16,7 @@ public class GmsStatusValueProvider {
 
   public String getGmsValue() {
     if (gmsValue == null || gmsValue.isEmpty()) {
-      gmsValue = AdNetworkUtils.isGooglePlayServicesAvailable(context) ? HAS_HGMS : NO_GMS;
+      gmsValue = AdNetworkUtils.isGooglePlayServicesAvailable(context) ? HAS_GMS : NO_GMS;
     }
     return gmsValue;
   }
--- a/app/src/main/java/cm/aptoide/pt/GmsStatusValueProvider.java
+++ b/app/src/main/java/cm/aptoide/pt/GmsStatusValueProvider.java
@@ -1,0 +1,23 @@
+package cm.aptoide.pt;
+
+import android.content.Context;
+import cm.aptoide.pt.dataprovider.ads.AdNetworkUtils;
+
+public class GmsStatusValueProvider {
+
+  private static final String HAS_HGMS = "Has GMS";
+  private static final String NO_GMS = "No GMS";
+  private Context context;
+  private String gmsValue = "";
+
+  public GmsStatusValueProvider(Context context) {
+    this.context = context;
+  }
+
+  public String getGmsValue() {
+    if (gmsValue == null || gmsValue.isEmpty()) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
```diff
@@ -334,12 +327,23 @@ private void setUTMDimensionsToUnknown() {
     FlurryAgent.addSessionProperty(UTM_CAMPAIGN, UNKNOWN);
     FlurryAgent.addSessionProperty(UTM_CONTENT, UNKNOWN);
     FlurryAgent.addSessionProperty(ENTRY_POINT, UNKNOWN);
+  }
+
+  private void sendRakamUserProperties(String utmContent, String utmSource, String utmCampaign,
+      String utmMedium, String entryPoint, String packageName) {
+    Identify rakamProperties = new Identify();
+
+    rakamProperties.set(GMS_RAKAM, rakamGmsValueCache);
```
jdandrade: rakamGmsValueCache might be null. It depends in another method being called, which is called in AnalyticsActivity... this should be reviewed.

app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
```diff
@@ -218,8 +219,12 @@ private void setupRakamFirstLaunchSuperProperty(boolean isFirstLaunch) {
   private void setupDimensions(android.app.Application application) {
     if (!checkForUTMFileInMetaINF(application)) {
       setUTMDimensionsToUnknown();
+      sendRakamUserProperties(UNKNOWN, UNKNOWN, UNKNOWN, UNKNOWN, UNKNOWN,
+          application.getPackageName());
     } else {
       setUserProperties(utmSource, utmMedium, utmCampaign, utmContent, entryPoint);
+      sendRakamUserProperties(utmContent, utmSource, utmCampaign, utmMedium, entryPoint,
+          application.getPackageName());
```
jdandrade: why don't we set them in the setUserProperties anyway? shouldn't the setUserProperties deal with all the user properties from different analytics?

jdandrade:
We are only reducing like 3 calls to 1. Shouldn't we check if we can send a batch of events with a single call? like we do for legacy BI?

app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
phillipsgoncalves: Created a wrapper that will get the value from the adnetworks (regarding gms). Started using it the first time i got it and will calculate it if it does not exist. I am using that "singleton" on both places.

app/src/main/java/cm/aptoide/pt/GmsStatusValueProvider.java
```diff
@@ -0,0 +1,23 @@
+package cm.aptoide.pt;
+
+import android.content.Context;
+import cm.aptoide.pt.dataprovider.ads.AdNetworkUtils;
+
+public class GmsStatusValueProvider {
+
+  private static final String HAS_HGMS = "Has GMS";
```
jdandrade: typo here?

app/src/main/java/cm/aptoide/pt/analytics/FirstLaunchAnalytics.java
phillipsgoncalves: I didnt want to merge both methods to keep rakam events sending independent

jdandrade: They can be independent, but UserProperties is a parent of RakamUserProperties, all names should be at the same level

PR #1215, opened 2020-01-02
[ASV-1846] fix: similar appc apps impression events 

jdandrade:
**What does this PR do?**

   The goal of PR is to solve the issue of "duplicated" impression events of the similar apps bundles in the app view when the user clicks install. What is currently happening in production is that when the user clicks install an app view with similar apps bundles, we send an similar appc and similar apps impression event. Then when the user scrolls down and has both bundles fully visible in the screen it would send a second impression event. 

  With this fix, we either send an event when user clicks install or when the user scrolls down to the fully visible bundle.

**Database changed?**

   No

- [ ] AppViewFragment.java

  Home > Open an App (with similar apps) > Click install > should send Appc_Similar_App_Interact impression event > scroll down > no more events sent.
  Home > Open an App (with similar apps) > scroll down until both similar apps bundles are fully visible > should send Appc_Similar_App_Interact impression event  (scrolling up and down again should not send)

  Tickets related to this pull-request: [ASV-1846](https://aptoide.atlassian.net/browse/ASV-1846)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
+++ b/app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
@@ -490,7 +490,6 @@ public class AppViewPresenter implements Presenter {
     view.getLifecycleEvent()
         .filter(event -> event.equals(View.LifecycleEvent.CREATE))
         .flatMap(lifecycleEvent -> Observable.merge(view.scrollVisibleSimilarApps()
-            .takeUntil(__ -> view.isSimilarAppsVisible())
             .map(__ -> true), view.similarAppsVisibilityFromInstallClick()))
         .first()
         .observeOn(Schedulers.io())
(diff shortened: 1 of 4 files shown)


app/src/main/java/cm/aptoide/pt/app/view/AppViewPresenter.java
```diff
@@ -501,11 +486,13 @@ private void sendSimilarAppsAdImpressionEvent(SimilarAppsViewModel similarAppsVi
...
 
-  private void handleOnScroll() {
+  private void handleOnSimilarAppsVisible() {
     view.getLifecycleEvent()
         .filter(event -> event.equals(View.LifecycleEvent.CREATE))
-        .flatMap(lifecycleEvent -> view.scrollVisibleSimilarApps())
-        .takeUntil(__ -> view.isSimilarAppsVisible())
+        .flatMap(lifecycleEvent -> Observable.merge(view.scrollVisibleSimilarApps()
+            .takeUntil(__ -> view.isSimilarAppsVisible())
+            .map(__ -> true), view.similarAppsVisibilityFromInstallClick()))
```
franciscoaleixo: Couldn't we simply do `Observable.merge(view.scrollVisibleSimilarApps(), view.similarAppsVisibilityFromInstallClick())`, since we have first() afterwards? Why takeUntil & map?

jdandrade: Yeah are right, will test and change it.

PR #1218, opened 2020-01-03
[ASV-2220] Read more detail view

phillipsgoncalves:
**What does this PR do?**

   This PR aims at fixing the dark theme in the read more description detail view inside AppView. Also took the chance to fix the toolbar inside this fragment as it was following a deprecated business rule. Before, we were setting the toolbar in read more with the color of the store where that same app was. Now we set the normal color (same as appview). This was discussed with Luis Guilherme.

**Database changed?**

 No

- [ ] DescriptionFragment.java

  Open the read more of several apps, and check that it has the normal color and never the color of the store where that app is.

  Tickets related to this pull-request: [ASV-2220](https://aptoide.atlassian.net/browse/ASV-2220)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/res/layout/fragment_app_view_description.xml
```diff
@@ -54,7 +52,6 @@
       android:layout_width="match_parent"
       android:layout_height="match_parent"
       android:layout_marginTop="?android:actionBarSize"
-      android:background="@color/white"
```
franciscoaleixo: Consider adding the background back to the same as app view

PR #1323, opened 2020-04-22
[MOB-538] Check available space

phillipsgoncalves:
**What does this PR do?**

   This PR aims at implementing a check to see if the device has enough space to install the app. Since the app after installing has occupies more space than the original apk, I added a buffer of 20%.

**Database changed?**

   No

- [ ] InstallAppSizeValidator.kt

On a device with low space, check that if you try to download a big app (example: creative destruction), you won't even start the download (you can check this by looking at Charles and by seeing that the progress bar won't be shown)

  Tickets related to this pull-request: [MOB-538](https://aptoide.atlassian.net/browse/MOB-538)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/ApplicationModule.java
+++ b/app/src/main/java/cm/aptoide/pt/ApplicationModule.java
@@ -182,6 +182,7 @@ import cm.aptoide.pt.home.bundles.ads.banner.BannerRepository;
 import cm.aptoide.pt.install.AppInstallerStatusReceiver;
 import cm.aptoide.pt.install.ForegroundManager;
 import cm.aptoide.pt.install.InstallAnalytics;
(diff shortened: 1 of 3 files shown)


franciscoaleixo:
We should check also the apk size (and not only the install size), i.e. (apksize + apksize*1.2) instead of only apksize*1.2

phillipsgoncalves:
While developing this ticket, several problems were brought up. João told me to simplify, so some things were not taken into account. Example: Updates. When you update an app you replace it, so you don't "all that new space".
Regarding this, due to this simplification, we are only checking for the size of the app after install, which means we are only checking the size of the data folder. The downloads of aptoide don't go to the "data" folder, they go to the ".aptoide".
I will talk with @jdandrade to see if this is necessary or not at the moment.

PR #1331, opened 2020-05-06
[bugfix/MOB-601] Out of space error bug

franciscoaleixo:
**What does this PR do?**

   Fixes a bug where if an user clicked install and an out of space error occured and then the user went back and re-entered the download view, the out of space error would appear without the user even clicking install.

   I'm not very much a fan of how it was eventually done, but my LOGAF is pretty low for this :) I think when we isolate downloads in general we can maybe think of a better solution. 

**Database changed?**

   No

  Ideally every fragment where there's downloads should be checked for the bugs presence. This is: app view, promotions view, editorial view, wallet install view, earn appc list view.

  [MOB-601](https://aptoide.atlassian.net/browse/MOB-601)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
+++ b/app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
@@ -318,13 +318,13 @@ public class AppViewFragment extends NavigationTrackFragment implements AppViewV
     super.onViewCreated(view, savedInstanceState);
 
     ViewStub.OnInflateListener installInflateListener = (viewStub, view1) -> {
-      install = ((Button) view1.findViewById(R.id.appview_install_button));
-      downloadInfoLayout = ((LinearLayout) view1.findViewById(R.id.appview_transfer_info));
-      downloadProgressBar = ((ProgressBar) view1.findViewById(R.id.appview_download_progress_bar));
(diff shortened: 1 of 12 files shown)


app/src/main/java/cm/aptoide/pt/app/view/AppViewFragment.java
```diff
@@ -318,13 +318,13 @@
     super.onViewCreated(view, savedInstanceState);
 
     ViewStub.OnInflateListener installInflateListener = (viewStub, view1) -> {
-      install = ((Button) view1.findViewById(R.id.appview_install_button));
```
franciscoaleixo, line 321: Most of these was somehow done when I cleaned up the code automatically 🤔

app/src/main/java/cm/aptoide/pt/promotions/PromotionsFragment.java
```diff
@@ -234,6 +229,14 @@ public void showPromotionApp(PromotionViewApp promotionViewApp, boolean isWallet
     promotionsView.setVisibility(View.VISIBLE);
   }
 
+  @Override public void showDownloadError(PromotionViewApp promotionViewApp) {
```
phillipsgoncalves, line 232: codestyle

PR #1334, opened 2020-05-07
[MOB-623] File Check

phillipsgoncalves:
**What does this PR do?**

This PR aims at adding the file check (see if files is already downloaded before checking if there is enough space to download)
This is only contemplating fully downloaded apps and not temporary ones.

**Database changed?**

   No

- [ ] InstalledAppSizeValidator.kt

Perform a download of any app. Then fill your device up (or on the getAvailablespace() method in the (InstalledAppSizeValidator.kt) class set it to return 0 and then (with the fully downloaded apk), try installing it again. Check if it just pops the install dialog instead of showing the no space dialog.

  Tickets related to this pull-request: [MOB-623](https://aptoide.atlassian.net/browse/MOB-623)

**Questions:**

**Code Review Checklist**

- [x] Codestyle
- [x] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/install/InstallAppSizeValidator.kt
+++ b/app/src/main/java/cm/aptoide/pt/install/InstallAppSizeValidator.kt
@@ -3,12 +3,17 @@ package cm.aptoide.pt.install
 import android.os.Build
 import android.os.Environment
 import android.os.StatFs
+import cm.aptoide.pt.database.realm.Download
+import cm.aptoide.pt.utils.FileUtils
 
-class InstallAppSizeValidator {
+class InstallAppSizeValidator(val filePathProvider: FilePathProvider) {
 
-  fun hasEnoughSpaceToInstallApp(downloadSize: Long): Boolean {
-    val availableSpace = getAvailableSpace()
-    return downloadSize <= availableSpace
+  fun hasEnoughSpaceToInstallApp(download: Download): Boolean {
+    if (isAppAlreadyDownloaded(download)) {
+      return true
+    } else {
+      return download.size <= getAvailableSpace()
+    }
   }
 
   private fun getAvailableSpace(): Long {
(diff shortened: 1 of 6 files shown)


app/src/main/java/cm/aptoide/pt/install/InstallAppSizeValidator.kt
```diff
@@ -3,12 +3,17 @@ package cm.aptoide.pt.install
...
 
-  fun hasEnoughSpaceToInstallApp(downloadSize: Long): Boolean {
-    val availableSpace = getAvailableSpace()
-    return downloadSize <= availableSpace
+  fun hasEnoughSpaceToInstallApp(download: Download): Boolean {
+    if (isAppAlreadyDownloaded(download)) {
+      return true
+    } else {
+      return download.size <= getAvailableSpace()
+    }
```
jdandrade, line 16: my logaf here is low but you could do something with the ? operator here like return a one liner, up to you tho.

app/src/main/java/cm/aptoide/pt/install/InstallAppSizeValidator.kt
```diff
@@ -20,4 +25,15 @@ class InstallAppSizeValidator {
...
 
+
+  private fun isAppAlreadyDownloaded(download: Download): Boolean {
+    for (fileToDownload in download.filesToDownload) {
+      if (!FileUtils.fileExists(
+              filePathProvider.getFilePathFromFileType(fileToDownload) + fileToDownload.fileName)) {
+        return false
+      }
+    }
+    return true
```
jdandrade, line 36: isn't returning true as a default here dangerous? if you would pass and empty filesToDownload list it would say true, maybe add a check on it at the beggining?

app/src/main/java/cm/aptoide/pt/install/InstallAppSizeValidator.kt
phillipsgoncalves: the ?: in kotlin is only for null check

"The Elvis operator in Kotlin is an operator that receives two inputs and returns the first argument if it is non-null or the second one otherwise. It is fancily called the null-coalescing operator. It is a variant of the ternary operator but for null-safety checking."

PR #1342, opened 2020-05-19
[HOTFIX/MOB-646] Install button refresh

phillipsgoncalves:
**What does this PR do?**

   This pr aims at fixing the button of the install button in appview after updating them. On André's phone, initially, any installation (update, downgrade, whatever) would not update the button, later on it was discovered that it was phone specific. After that i was able to discover that i could reproduce that behavior by downgrading an app and then trying to update it. Then, after the update was installed, the button would not update to open because the update from the db would be removed in the downgrade and then when we would try to get it on the database replaced (update) call, it would not exist so the db would not emit (Room does not reply wiht an empty list like Realm, BUT in case the method is single it will throw an exception that we can map to null).

**Database changed?**

   No

- [ ] InstalledIntentService.java

Downgrade an app and then try to update it. Check if the button updates well. Also, try to do other type of installations to confirm everything is ok, like update apps, downgrade, install, etc.

  Tickets related to this pull-request: [MOB-646](https://aptoide.atlassian.net/browse/MOB-646)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/home/apps/AppsManager.java
+++ b/app/src/main/java/cm/aptoide/pt/home/apps/AppsManager.java
@@ -276,7 +276,8 @@ public class AppsManager {
               return Single.just(value);
             })
             .flatMapCompletable(download -> installManager.install(download))
-            .andThen(aptoideInstallManager.sendConversionEvent()));
+            .andThen(aptoideInstallManager.sendConversionEvent()))
+        .onErrorComplete();
   }
 
   public boolean showWarning() {
--- a/app/src/main/java/cm/aptoide/pt/home/apps/AppsManager.java
+++ b/app/src/main/java/cm/aptoide/pt/home/apps/AppsManager.java
@@ -276,8 +276,7 @@ public class AppsManager {
               return Single.just(value);
             })
             .flatMapCompletable(download -> installManager.install(download))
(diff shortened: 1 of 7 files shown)


app/src/main/java/cm/aptoide/pt/home/apps/AppsManager.java
```diff
@@ -276,8 +276,7 @@ public Completable updateApp(App app, boolean isAppcUpdate) {
               return Single.just(value);
             })
             .flatMapCompletable(download -> installManager.install(download))
-            .andThen(aptoideInstallManager.sendConversionEvent()))
-        .toCompletable();
+            .andThen(aptoideInstallManager.sendConversionEvent()));
```
phillipsgoncalves, line 279: @franciscoaleixo i have some doubts here. This chain will get the update from the db in order to create the download. Before, since it was an observable and if the update did nothing would happen. Now since it is a single, it will return null, which might lead to a NP.
Technically, for it to happen would be really weird, because the app would have to be on the list of updates, but the update would not be in the database. I am not seeing how this would happen, so it is a minor problem, but i still think we should try to defend ourselves from these possible mistakes. Currently it will go for the doOnError and log the error. Since it has a retry around the button click, nothing would happen, just like before.

PR #1362, opened 2020-06-24
Feature/mob 657 updates notification abtest

paquiquinho:
**What does this PR do?**

This PR adds a new system to handle updates notifications and adds an abtest to the flow to change rules and design;

**Database changed?**

No

- [ ] UpdatesNotificationManager.kt

Force each group in the code and test all conditions described in the ticket for each. 

NOTES:
- I recommend setting the repeat time from 1 day to 15 minutes (lowest possible time, there might be a variable in Work for this in milis);
- If you have no new updates since the last time the work was fired, the work will trigger and just do nothing, having to wait another period to trigger again;
- The period you set for repeating is inexact, i.e.: if you set it to 15 minutes, it might take 12 or it might take 25. This happens because of android system management. From the documentation I could find it should never exceed two periods and from my tests, it hardly ever exceeds 1/3 of the period (over 20 min in my 15 min tests);

  Tickets related to this pull-request: [MOB-657](https://aptoide.atlassian.net/browse/MOB-657)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


phillipsgoncalves:
We spoke about it, but just leaving it for the status check. Tested some of the notifications and it is working fine except the 1st one (after setting up the worker). It is only coming up when i open the app, even though i installed some new apps (older versions and followed those stores) after it registered the worker

PR #1365, opened 2020-06-25
[feature/MOB-656] Bonus appc flair

franciscoaleixo:
**What does this PR do?**

   This PR implements the new bonus appc flair in AppView and the new featured appcoins bundle. There was a major change that really has to be reviewed which is that the base theme now extends from MaterialComponents instead of AppCompat. This is especially needed in order to have the border in the flair, but there's consequences from that: every non-specific XML tag now extends from material components (e.g. <Button> is now the button from material library instead of AppCompat, so it needed to be replaced with <android.widget.Button>.

**NOTE:** Strings are still hardcoded.

**Database changed?**

   No

  Check if the new bundle appears correctly and the flair correctly appears in AppView on apps with appcoins billing (check Zeplin to see if it matches the design). You can also mock the bonus response to check how it behaves if there's no bonus (the old 'Featured AppCoins Apps' bundle should appear and no flair should appear on AppView).

  Tickets related to this pull-request: [MOB-XXXX](https://aptoide.atlassian.net/browse/MOB-XXXX)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/home/bundles/appcoins/FeaturedAppcBundleAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+import rx.subjects.PublishSubject
+import java.text.DecimalFormat
+
+class FeaturedAppcBundleAdapter(var apps: List<Application>,
+                                val oneDecimalFormatter: DecimalFormat,
+                                val appClickedEvents: PublishSubject<HomeEvent>) :
+    RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+
+  private var homeBundle: HomeBundle? = null
+  private var bundlePosition: Int = -1
```
phillipsgoncalves, line 21: Is it safe to use this position variable? I know there is not too much to do as the AppsBundleViewHolder receives the position..

PR #1372, opened 2020-07-15
[feature/MOB-785 & MOB-784] Search Filters

franciscoaleixo:
**What does this PR do?**

   This PR Implements the new search filters. This also includes thew new empty state, MOB-784.

**Database changed?**

   No

  Check general search and specific store search. Both should work correctly and as expected. Test the different the filters (and different combinations) and check if they're working as expected (also in older devices). 

  [MOB-785](https://aptoide.atlassian.net/browse/MOB-785)
  [MOB-784](https://aptoide.atlassian.net/browse/MOB-784)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


dataprovider/src/main/java/cm/aptoide/pt/dataprovider/ws/v7/ListSearchAppsRequest.java
```diff
@@ -26,6 +26,8 @@
  */
 public class ListSearchAppsRequest extends V7<ListSearchApps, ListSearchAppsRequest.Body> {
 
+  private static final int LIMIT = 15;
```
jdandrade, line 29: Can't remember if we decided changing this, it was updated from 10->15 right?

app/src/main/java/cm/aptoide/pt/search/SearchManager.java
```diff
@@ -59,76 +51,52 @@ public SearchManager(SharedPreferences sharedPreferences, TokenInvalidator token
...
+            matureEnabled));
+  }
+
+  public SearchFilters getSearchFilters(List<Filter> viewFilters) {
+    boolean onlyFollowedStores = false;
+    boolean onlyTrustedApps = false;
+    boolean onlyBetaApps = false;
+    boolean onlyAppcApps = false;
+    for (Filter filter : viewFilters) {
+      if (filter.getIdentifier() == null) continue;
```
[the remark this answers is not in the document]
franciscoaleixo: In this case no, I simply made it nullable because technically the identifier is not needed to have a fully functional view of filters (e.g. if for example every filter were to function the same way). I don't have a particular scenario for this though so I could remove it

dataprovider/src/main/java/cm/aptoide/pt/dataprovider/ws/v7/ListSearchAppsRequest.java
franciscoaleixo: I think initially it was decided 25, but after the decision to not have local filters, I figured I'd drop it down to 15 because when I was testing 25, it seemed too much of a padding and too much apps in one call for nothing.

So yeah, I changed it from 10 -> 15

jdandrade: ok, we can leave 15.

PR #1383, opened 2020-08-04
[bugfix/MOB-861] Follow Apps and Catappult by default

franciscoaleixo:
**What does this PR do?**

   When changing from Realm to Room, there's an issue that if an user upgrades from a version that has Room to a version with Realm, the user stops following the default stores. This fixes it and generalizes the case of updating the app to a specific version.

**Database changed?**

   No

  Install an older version with realm, e.g. 9.13.3.0, and then install this version (it has to count as an update, so it needs to have the same signature). Check that it follows the default stores.

  [MOB-861](https://aptoide.atlassian.net/browse/MOB-861)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/FirstLaunchManager.kt
```diff
@@ -0,0 +1,81 @@
...
+                         private val rootAvailabilityManager: RootAvailabilityManager,
+                         private val accountManager: AptoideAccountManager,
+                         private val shortcutManager: AptoideShortcutManager,
+                         private val context: Context) {
+
+  fun runFirstLaunch(): Completable {
+    return Completable.mergeDelayError(setSharedPreferencesValues(), generateAptoideUuid())
+        .andThen(Completable.mergeDelayError(followedStoresManager.setDefaultFollowedStores(),
+            updateRepository.sync(true, false),
+            rootAvailabilityManager.updateRootAvailability(), accountManager.updateAccount(),
```
phillipsgoncalves, line 32: Before updateAccount and createshortcut were both inside a merge. To be honest, I dont know why, but does it affect this ? Would it just need of them to complete ?

app/src/main/java/cm/aptoide/pt/FirstLaunchManager.kt
```diff
@@ -0,0 +1,81 @@
...
+            createShortcut()))
+        .doOnError { e -> CrashReport.getInstance().log(e) }
+  }
+
+  private fun generateAptoideUuid(): Completable {
+    return Completable.fromAction { idsRepository.getUniqueIdentifier() }
```
phillipsgoncalves, line 38: Android studio is giving a warning to use property access syntax

app/src/main/java/cm/aptoide/pt/UpdateLaunchManager.kt
```diff
@@ -0,0 +1,17 @@
+package cm.aptoide.pt
+
+import rx.Completable
+
+class UpdateLaunchManager(val followedStoresManager: FollowedStoresManager) {
+
```
phillipsgoncalves, line 6: Current version code not being used

franciscoaleixo: I can remove but I did it on purpose because there might be specific scenarios where you only want to do a change coming from a specific version to another specific version. I think it's generally standard to pass the previous and the current version in these type of things.

app/src/main/java/cm/aptoide/pt/FirstLaunchManager.kt
franciscoaleixo: Personally, I don't like to use the property access syntax in methods where it's not a simple getter (i.e. it doesn't just return a field). Which is this case, which is why I didn't use it.

app/src/main/java/cm/aptoide/pt/FirstLaunchManager.kt
franciscoaleixo: As we've discussed, no. The merge waits for every completable to complete.

PR #1393, opened 2020-08-27
[MOB-964] ToolboxContentProvider changes

danfbfern:
**What does this PR do?**
Adds two fields necessary for the new AutoLogin uploader view.
Name (nickname) and the Avatar path.

**Database changed?**
 No

- ToolboxContentProvider.java

Using an uploader version with the autologin changes, login in vanilla and then go to uploader and check if the name and avatar show up.

https://aptoide.atlassian.net/browse/MOB-964

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/toolbox/ToolboxContentProvider.java
+++ b/app/src/main/java/cm/aptoide/pt/toolbox/ToolboxContentProvider.java
@@ -123,7 +123,7 @@ public class ToolboxContentProvider extends ContentProvider {
           return create("loginAvatar", account.getAvatar());
         default:
           throw new IllegalArgumentException(
-              "Only /token, /repo, /passHash, /loginType and /loginName supported.");
+              "Only /token, /refreshToken, /repo, /passHash, /loginType, /loginName, loginNickname and loginAvatar supported.");
       }
     } else {
       throw new SecurityException("Package not authorized to access provider.");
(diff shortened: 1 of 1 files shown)


app/src/main/java/cm/aptoide/pt/toolbox/ToolboxContentProvider.java
```diff
@@ -62,11 +64,13 @@
     uriMatcher = new UriMatcher(UriMatcher.NO_MATCH);
     uriMatcher.addURI(BuildConfig.CONTENT_AUTHORITY, "token", TOKEN);
     uriMatcher.addURI(BuildConfig.CONTENT_AUTHORITY, "refreshToken", REFRESH_TOKEN);
-    uriMatcher.addURI(BuildConfig.CONTENT_AUTHORITY, "repo", REPO);
+    uriMatcher.addURI(BuildConfig.CONTENT_AUTHORITY, "storeName", STORE_NAME);
```
jdandrade, line 67: warning: by doing this change you will break older versions of uploader and also other potential clients that use this (backup-apps, toolbox)
Because we don't have a way to version this right now in place, it is better to only make additions

jdandrade (changes requested):

app/src/main/java/cm/aptoide/pt/toolbox/ToolboxContentProvider.java
danfbfern: I only changed it because of this comment https://github.com/Aptoide/aptoide-uploader/pull/37#discussion_r480980865
I can revert it again, the main reason i kept the same names as before was to avoid problems, but based on the comment I assumed I supposed to change it anyway

jdandrade (approved):

PR #1399, opened 2020-09-18
[bugfix/MOB-478] Notification Ready To Install (Android 10-11)

franciscoaleixo:
**What does this PR do?**

   This implements a new notification that appears when a download is finished in background to warn users that their app is ready. It uses the default notification system that we have with Aptoide but it uses a specific channel and it has a few quirks. One of those is the fact that we have to track if the notification is shown (manually) to change the type of notification we show. See the testing section to undestand the constraints.

**NOTE:** Strings are still hardcoded and must be fixed!

**Database changed?**

   No

  This PR implements two similar notifications that has to be tested. The constraints are as follows:
    - Both notifications will only appear when a download is finished and the app is in background (is either "minimized" or closed")
    - Both notifications will only appear for devices with Android 10 or newer (11 for now).

  There's two types of notifications:
    - Single app notification: This is a notification to warn the user that a specific app has finished downloading. It has the app name and icon. Clicking it will redirect to its appview and start the installation immediately.
    - Multi-app notification: This is a notification that only appears if another download is finished WHILE there's a single app notification (or multi) already active. If there is, the single app notification will transform to a multi-app notification. Clicking this will redirect to the Apps section. This is especially useful notification for "update all" in Apps.

  [MOB-478](https://aptoide.atlassian.net/browse/MOB-478)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/view/DeepLinkManager.java
```diff
@@ -343,6 +376,11 @@ private void fromHomeDeepLink() {
     bottomNavigationNavigator.navigateToHome();
   }
 
+  private void goToAppsDownloadsSection() {
+    // TODO: Scroll to Downloads
```
phillipsgoncalves, line 380: todo ?

app/src/main/java/cm/aptoide/pt/view/DeepLinkManager.java
```diff
@@ -263,10 +283,23 @@ private void appViewDeepLink(long appId, String packageName, boolean showPopup,
...
-        : AppViewFragment.OpenType.OPEN_ONLY;
-    appNavigator.navigateWithPackageAndStoreNames(packageName, storeName, openType);
+  private void appViewDeepLink(String packageName, String storeName, String openType) {
+    appNavigator.navigateWithPackageAndStoreNames(packageName, storeName, getOpenType(openType));
+  }
+
+  private AppViewFragment.OpenType getOpenType(String openType) {
+    if (openType == null) return AppViewFragment.OpenType.OPEN_ONLY;
+    switch (openType) {
+      case "open_and_install":
```
phillipsgoncalves, line 293: shouldnt these be a field ? So that it is less error prone while trying to set it to use it?

phillipsgoncalves (changes requested):

phillipsgoncalves:
Soon will post trhe results of the tests made

phillipsgoncalves:
Note: When reviewing the PR I noticed that sometimes, the notification of a 2nd downloaded app would not disappear from the downloading state and we had a crash related to the DownloadNotFoundException which would happen if the download is null or if the files associated to the download were not found in a method that was only being called when we try to trigger the installation.
I also noticed another problem, which was considered not major, where due to multiple emissions from the InstallManager, when we do two downloads, while trying to setup the notification for the 2nd downloaded app, we will setup the another notification for the 1st downloaded app.

PR #1425, opened 2020-11-26
[APP-81] Fix out of space inconsistency

phillipsgoncalves:
**What does this PR do?**

This PR aims at setting the download to the error state when it encounters an error while trying to move the download files. We are setting the general error because from the exception we can not distinguish the different exceptions.

Please confirm that the everything is working correctly

**Database changed?**

No

- [ ] AptoideDownloadManager.java

On a low storage device try to perform a download that only fails due to storage at the end of the download. For example, I tested downloading Lords Mobile on an emulator with 1000MB of internal storage and no SD Card.
This leads do the weird state that we were getting where it is possible do start the download, but it is not possible to move the file (this happens because when we are moving it, it clones the file).

  Tickets related to this pull-request: [APP-81](https://aptoide.atlassian.net/browse/APP-81)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


jdandrade (approved):
lgtm

PR #1427, opened 2020-11-27
[Feature/APP-103] new setting: enforce native installer

jdandrade:
**What does this PR do?**

   new setting - enforce native installer. This will allow Xiaomi users to recover from the state of not being able to install any application in the app because of the PackageInstaller.

EDIT: Strings on this branch are on temporary. The ones from dev are the correct ones.

**Database changed?**

  No

- [ ] settings.xml
- [ ] AppBundlesVisibilityManager.java
- [ ] InstallManager.java

  - Test installations with both setting on/off on Xiaomi and non-xiaomi devices.

  Tickets related to this pull-request: [APP-103](https://aptoide.atlassian.net/browse/APP-103)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/install/InstallManager.java
+++ b/app/src/main/java/cm/aptoide/pt/install/InstallManager.java
@@ -18,6 +18,7 @@ import cm.aptoide.pt.downloadmanager.DownloadsRepository;
 import cm.aptoide.pt.install.installer.InstallCandidate;
 import cm.aptoide.pt.install.installer.InstallationState;
 import cm.aptoide.pt.logger.Logger;
+import cm.aptoide.pt.preferences.managed.ManagedKeys;
 import cm.aptoide.pt.preferences.managed.ManagerPreferences;
 import cm.aptoide.pt.preferences.secure.SecurePreferences;
 import cm.aptoide.pt.root.RootAvailabilityManager;
@@ -240,11 +241,15 @@ public class InstallManager {
   }
 
   public Completable install(RoomDownload download) {
-    return install(download, false, false, true);
+    boolean shouldForceDefaultInstall =
+        sharedPreferences.getBoolean(ManagedKeys.ENFORCE_NATIVE_INSTALLER_KEY, false);
+    return install(download, shouldForceDefaultInstall, false, true);
   }
 
   public Completable install(RoomDownload download, boolean shouldInstall) {
-    return install(download, false, false, shouldInstall);
+    boolean shouldForceDefaultInstall =
+        sharedPreferences.getBoolean(ManagedKeys.ENFORCE_NATIVE_INSTALLER_KEY, false);
+    return install(download, shouldForceDefaultInstall, false, shouldInstall);
   }
 
   private Completable defaultInstall(RoomDownload download) {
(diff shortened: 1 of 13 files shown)


app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -240,19 +241,25 @@ private boolean isAppInstalling(List<RoomInstalled> installingAppList, String pa
   }
 
   public Completable install(RoomDownload download) {
-    return install(download, false, false, true);
+    boolean shouldForceDefaultInstall =
```
phillipsgoncalves, line 244: Why not use the settingsProvider you created here? You have a method for isEnforceNativeInstaller() and you wouldn't need to extract the variable either

app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -240,19 +241,25 @@ private boolean isAppInstalling(List<RoomInstalled> installingAppList, String pa
...
+        sharedPreferences.getBoolean(ManagedKeys.ENFORCE_NATIVE_INSTALLER_KEY, false);
+    return install(download, shouldForceDefaultInstall, false, true);
   }
 
   public Completable install(RoomDownload download, boolean shouldInstall) {
-    return install(download, false, false, shouldInstall);
+    boolean shouldForceDefaultInstall =
+        sharedPreferences.getBoolean(ManagedKeys.ENFORCE_NATIVE_INSTALLER_KEY, false);
```
phillipsgoncalves, line 251: same as above

app/src/main/java/cm/aptoide/pt/install/InstallManager.java
```diff
@@ -240,19 +241,25 @@ private boolean isAppInstalling(List<RoomInstalled> installingAppList, String pa
...
   }
 
   private Completable defaultInstall(RoomDownload download) {
     return install(download, true, false, true);
   }
 
   public Completable splitInstall(RoomDownload download) {
-    return install(download, false, true, true);
+    boolean shouldForceDefaultInstall =
+        sharedPreferences.getBoolean(ManagedKeys.ENFORCE_NATIVE_INSTALLER_KEY, false);
```
phillipsgoncalves, line 261: same as above

phillipsgoncalves: Also here, note for the future, but forceSplitInstall and forceDefaultInstall are mutually exclusive as we can only use one type of install for each install.

phillipsgoncalves (changes requested):

phillipsgoncalves:
After some tests, I noticed the following:
- after changing the installer setting should we refresh the updates automatically? This is important because after changing the setting, if we navigate to the apps section, we will still have old updates/updates with app bundles which will not be installed.
- some updates are stuck on installing even after the installation finished (wether with app not installed or success) (this happened on my pixel 5, could not reproduce on my xiami as on my xiaomi device i could not see those apps)

jdandrade:
> * after changing the installer setting should we refresh the updates automatically? This is important because after changing the setting, if we navigate to the apps section, we will still have old updates/updates with app bundles which will not be installed.

Good question. This feature is to be a fallback to a really bad experience, a lifeboat for users stuck on this situation. Even if the UX is not optimal this users will most likely be in contact with us about the issue so we can just tell them to refresh/clear data on the device.

PR #1449, opened 2021-04-29
Bug 125527

osalamakhagl:
Fixed Aptoide crash after installing the application


apkbuilder-lw:
Can one of the admins verify this patch?

PR #1504, opened 2021-11-02
[APP-431] Promotional Cards

phillipsgoncalves:
**What does this PR do?**

This PR aims at implementing the new promotional cards card, coming soon card.
In this PR I also fixed "home_interact" (impression and click) events for all promotional cards. 

**Database changed?**

   No

- [ ] Home.java

For the analytics part, check that they are being correctly sent (impression and click on the respective items).
Regarding the promotional card, ask me for the rewrites so that you can see the cards. Then check that the card is correctly mapped, press "notify me" and check that you receive the notification for the app that was requested.

  Tickets related to this pull-request: [APP-431](https://aptoide.atlassian.net/browse/APP-431)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/notification/ComingSoonNotificationWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+        .doOnSuccess { detailedAppResult: DetailedAppRequestResult? ->
+          if (detailedAppResult != null && detailedAppResult.detailedApp != null) {
+            cancelComingSoonVerification(packageName)
+            handleAppArrived(ComingSoonApp(
+                detailedAppResult.detailedApp.name,
+                detailedAppResult.detailedApp.icon, detailedAppResult.detailedApp.md5,
+                detailedAppResult.detailedApp.store.name, detailedAppResult.detailedApp.packageName
+            ))
+          }
+        }.toBlocking().value()
```
jdandrade, line 37: We probably want to avoid continuing to propagate this practice in background operations of retrieving data from an observable datasource with toBlocking(). 
Here we can work this around using the androidx.work:work-rxjava2 lib by the sdk team that gives us an observable interface for doWork(). Just need to extend RxWorker instead of Worker. That way we should be able to avoid the toBlocking().value() and dealing with errors if necessary (probably here we just want logs).

app/src/main/java/cm/aptoide/pt/notification/ComingSoonNotificationWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import androidx.work.WorkManager
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import cm.aptoide.pt.AptoideApplication
+import cm.aptoide.pt.DeepLinkIntentReceiver
+import cm.aptoide.pt.R
+import cm.aptoide.pt.app.aptoideinstall.ComingSoonApp
+import cm.aptoide.pt.logger.Logger
```
jdandrade, line 15: This logger is not being used.

app/src/main/java/cm/aptoide/pt/notification/ComingSoonNotificationManager.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    WorkManager.getInstance(context)
+        .enqueueUniquePeriodicWork(
+            WORKER_TAG + url, ExistingPeriodicWorkPolicy.KEEP,
+            comingSoonWorkRequest)
+  }
+
+  private fun setUpChannel() {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      val name = "Coming Soon notification"
```
jdandrade, line 48: Use the same pattern we are using on other places: "Coming Soon notifications"

app/src/main/java/cm/aptoide/pt/notification/ComingSoonNotificationWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+import cm.aptoide.pt.networking.image.ImageLoader
+import cm.aptoide.pt.view.app.AppCenter
+import cm.aptoide.pt.view.app.DetailedAppRequestResult
+
+class ComingSoonNotificationWorker(private val context: Context,
+                                   private val workerParameters: WorkerParameters,
```
jdandrade, line 21: can remove the val property since this is not being used inside the class

app/src/main/java/cm/aptoide/pt/home/bundles/BundlesResponseMapper.java
```diff
@@ -290,8 +294,11 @@ private ActionItem map(ActionItemResponse viewObject) {
           if (widget.getType()
               .equals(Type.NEWS_ITEM)) {
             return HomeBundle.BundleType.NEWS_ITEM;
-          } else {
+          } else if (widget.getType()
+              .equals(Type.IN_GAME_EVENT)) {
             return HomeBundle.BundleType.IN_GAME_EVENT;
+          } else {
+            return HomeBundle.BundleType.APP_COMING_SOON;
```
jdandrade, line 301: This might be an error that was already in prod. We should return APP_COMING_SOON only if the widget type is of that type. Return UNKNOWN on else. I'm not sure what is the current behavior (if this version in the future receives new types of PROMO bundles)

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -717,4 +723,27 @@ private void handleLoadMoreError() {
...
+    view.getLifecycleEvent()
+        .filter(lifecycleEvent -> lifecycleEvent.equals(View.LifecycleEvent.CREATE))
+        .flatMap(created -> view.notifyMeClicked())
+        .filter(homeEvent -> homeEvent.getBundle() instanceof ActionBundle)
+        .doOnNext(event -> {
+          ActionBundle bundle = (ActionBundle) event.getBundle();
+          homeAnalytics.sendPromotionalArticleClickEvent(bundle.getType()
+              .name(), bundle.getActionItem()
+              .getCardId());
+          homeAnalytics.sendActionItemEditorialTapOnCardInteractEvent(bundle.getTag(),
```
jdandrade, line 737: As we talked we can probably remove the Editorial from this method name since it is no longer strict to editorial.

jdandrade (changes requested):

PR #1516, opened 2021-12-22
[APP-776] Notify me

phillipsgoncalves:
**What does this PR do?**

This PR aims at implementing the cancel flow for the notify me button on the app coming soon card.

**Database changed?**

   No

- [ ] BundlesResponseMapper.java

Press the Notify me button, check that you receive the notification. Repeat the flow, but press the cancel button after pressing the notify me button and check that you DO NOT receive the notification.

  Tickets related to this pull-request: [APP-776](https://aptoide.atlassian.net/browse/APP-776)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -704,9 +708,43 @@ private void handleLoadMoreError() {
                   .getCardId());
         })
         .map(HomeEvent::getBundle)
-        .cast(ActionBundle.class)
-        .flatMapCompletable(bundle -> home.setupAppComingSoonNotification(bundle.getActionItem()
-            .getUrl()))
+        .cast(AppComingSoonPromotionalBundle.class)
+        .flatMap(bundle -> home.setupAppComingSoonNotification(bundle.getActionItem()
+            .getUrl())
+            .andThen(Observable.just(bundle)))
```
jdandrade, line 714: This is minor and up 2 you but I would change this to not force an observable creation with bundle emission when it looks like you want to actually map the emission back to bundle.

```suggestion
        .flatMap(bundle -> home.setupAppComingSoonNotification(bundle.getActionItem()
            .getUrl())
            .toObservable().map(__ -> bundle))
```

We actually want to subscribe to a fire and forget completable without changing the object... like a doOnNext that takes a function... But I don't even know the best way to do that. doOnNext with a completable and .subscribe there is just calling for garbage collection problems... dunno :hammer:

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -704,9 +708,43 @@ private void handleLoadMoreError() {
...
+        .doOnNext(bundle -> {
+          bundle.setRegisteredForNotification(true);
```
jdandrade, line 716: I don't agree with this manipulation here. There are 3 entities currently manipulating this field:
- The repo/service which creates the object with the default (false)
- This presenter which is changing it for the notify click and cancel notify click.
- The adapter to update the bundle in the view.

We could at least just pass the value down to the adapter and cut the 2nd manipulation which seems a bit unnecessary in my opinion.

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -704,9 +708,43 @@ private void handleLoadMoreError() {
...
+        })
+        .map(HomeEvent::getBundle)
+        .cast(AppComingSoonPromotionalBundle.class)
+        .flatMap(bundle -> home.cancelAppComingSoonNotification(bundle.getActionItem()
+            .getUrl())
+            .andThen(Observable.just(bundle)))
+        .doOnNext(homeBundle -> {
+          homeBundle.setRegisteredForNotification(false);
+          view.updateAppComingSoonStatus(homeBundle);
+        })
```
jdandrade, line 747: pretty much the exact same comments as the chain above.

app/src/main/java/cm/aptoide/pt/home/bundles/BundlesAdapter.java
```diff
@@ -296,4 +297,21 @@ public boolean isLoaded() {
...
+            .getUrl()
+            .equals(homeBundle.getActionItem()
+                .getUrl())) {
+          int index = bundles.indexOf(bundle);
+          ((AppComingSoonPromotionalBundle) bundle).setRegisteredForNotification(
+              homeBundle.isRegisteredForNotification());
+          notifyItemChanged(index);
+        }
+      }
+    }
```
jdandrade, line 315: I may be oversimplifying but... can't this be replaced by?:

```suggestion
    int index = bundles.indexOf(homeBundle);
    ((AppComingSoonPromotionalBundle) bundles.get(index)).setRegisteredForNotification(homeBundle.isRegisteredForNotification());
    notifyItemChanged(index);
```

app/src/main/java/cm/aptoide/pt/notification/ComingSoonNotificationManager.kt
```diff
@@ -5,10 +5,12 @@ import android.app.NotificationManager
 import android.content.Context
 import android.os.Build
 import androidx.work.*
+import cm.aptoide.pt.home.AppComingSoonPreferencesManager
 import rx.Completable
 import java.util.concurrent.TimeUnit
 
-class ComingSoonNotificationManager(private val context: Context) {
+class ComingSoonNotificationManager(private val context: Context,
```
jdandrade, line 12: This is related to last PR but ideally this class should not have a context and depend on the instance of workmanager and notificationmanager.

jdandrade (changes requested):

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
```diff
@@ -704,9 +708,43 @@ private void handleLoadMoreError() {
                   .getCardId());
         })
         .map(HomeEvent::getBundle)
-        .cast(ActionBundle.class)
-        .flatMapCompletable(bundle -> home.setupAppComingSoonNotification(bundle.getActionItem()
-            .getUrl()))
+        .cast(AppComingSoonPromotionalBundle.class)
```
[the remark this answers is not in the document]
phillipsgoncalves: I did it, but dont know if it damages the readability, which is the reason why a lot of times i dont like using pairs.

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
phillipsgoncalves: Tested this and as we spoke on the office, it did not work. The setupNotification method is completable, so it is a completed observable, meaning we can not use doOnNext after, which is what we need.

app/src/main/java/cm/aptoide/pt/home/HomePresenter.java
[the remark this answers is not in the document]
jdandrade: We are still doing the double cast.

jdandrade (changes requested):
Missing migration. Log:
    java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number.

jdandrade (approved):
lgtm

PR #1542, opened 2022-07-11
[APP-733] size calculation error

phillipsgoncalves:
**What does this PR do?**

This PR aims at fixing the bug we had regarding the download percentage.
This bug was happening only on apps with PAD and PFD. The reason for it was that we were only adding the main dynamic split apk size, instead of also summing the config apk's size of those dynamic split apks.

Also took the oportunity to count that size also for the updates, which had not been done previously

**Database changed?**

   No

- [ ] DownloadFactory.java

Download an app (for example wwe supercard or league of legends: wild rift) and check that everything is ok regarding the app size and download percentage.

  Tickets related to this pull-request: [APP-733](https://aptoide.atlassian.net/browse/APP-733)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/download/DownloadFactory.java
+++ b/app/src/main/java/cm/aptoide/pt/download/DownloadFactory.java
@@ -251,6 +251,9 @@ public class DownloadFactory {
     long dynamicSplitsSize = 0;
     for (DynamicSplit dynamicSplit : dynamicSplits) {
       dynamicSplitsSize += dynamicSplit.getFileSize();
+      for (Split configSplit : dynamicSplit.getConfigSplits()) {
+        dynamicSplitsSize += configSplit.getFilesize();
+      }
     }
     return dynamicSplitsSize + appBaseSize;
   }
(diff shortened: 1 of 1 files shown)


app/src/main/java/cm/aptoide/pt/download/DownloadFactory.java
```diff
@@ -250,7 +250,15 @@ public RoomDownload create(int downloadAction, String appName, String packageNam
   private long calculateAppSize(long appBaseSize, List<DynamicSplit> dynamicSplits) {
     long dynamicSplitsSize = 0;
     for (DynamicSplit dynamicSplit : dynamicSplits) {
-      dynamicSplitsSize += dynamicSplit.getFileSize();
+
+      if (dynamicSplit.getDeliveryTypes()
+          .contains("INSTALL_TIME")) {
+
+        dynamicSplitsSize += dynamicSplit.getFileSize();
```
jdandrade, line 257: won't this just add size if the split is install_time?

jdandrade (changes requested):

jdandrade (approved):

PR #1582, opened 2022-11-10
[DT-21] implement install manager for DT

G0retZ:
**What does this PR do?**

   Adds InstallManager implementation for DT service layer.

**Database changed?**

   Yes

- Better by commits

  This is not connected to UI, so essentially just run the app and make sure no new errors in logs

  Tickets related to this pull-request: [DT-21](https://aptoide.atlassian.net/browse/DT-21)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass


install-manager/src/main/java/cm/aptoide/pt/install_manager/App.kt
```diff
@@ -17,7 +17,7 @@ import cm.aptoide.pt.install_manager.dto.InstallPackageInfo
 interface App<D> {
   val packageName: String
   val installedVersionName: String?
-  val installedVersionCode: Int?
```
jdandrade, line 20: What is the reason for this change? I believe that an apps version code according to google documentation is a positive integer, meaning we don't actually need a Long value here? Just curious.

jdandrade (approved):

install-manager/src/main/java/cm/aptoide/pt/install_manager/App.kt
G0retZ: Google changed the API for this since SDK `28` now it is Long.
https://developer.android.com/reference/android/content/pm/PackageInfo#getLongVersionCode()

jdandrade: Good to know, makes sense, thanks.

PR #1725, opened 2023-11-24
[DT 902] Payment with credit card fails on second attempt when first attempt is cancelled

ricardo-frango:
**What does this PR do?**

Fixes an issue of an unsuccessful buy attempt after a first cancelled attempt

**Database changed?**

   No

**Steps to reproduce
1º - Go to credit card payment
2º - Introduce a card with redirect
3º - On the webview, cancel
4º - Go to credit card payment again
5º - Introduce the same card
6º - On the webview add the password to pay
7º - Failed but should be success

  Flow on how to test this or QA Tickets related to this use-case: [APP-902](https://aptoide.atlassian.net/browse/DT-902)

  Tickets related to this pull-request: [APP-902](https://aptoide.atlassian.net/browse/DT-902)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/payments/payment-methods/adyen/src/main/java/cm/aptoide/pt/payment_method/adyen/presentation/AdyenCreditCardViewModel.kt
+++ b/payments/payment-methods/adyen/src/main/java/cm/aptoide/pt/payment_method/adyen/presentation/AdyenCreditCardViewModel.kt
@@ -258,6 +258,7 @@ class AdyenCreditCardViewModel(
         action = action,
         configuration = redirectConfiguration
       ) { actionData ->
+        // FIX: this should not be called on second attempt with the data of the previous buy attempt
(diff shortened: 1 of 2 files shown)


ricardo-frango:
I was not able to found a solution to this because it was hard to understand what was happening.
What is happening is that on the first attempt to buy, when we cancel the Adyen redirect component saves the data even when the webview returns cancel.
what happens on the second attempt is that the callback is called right before calling the webview. this causes the transactions endpoint to be called and the data of the second buy attempt is mixed with the redirect data of the first buy attempt. this is probably causing the data to be invalidated which causes the buy failure on the second attempt.
I left a branch and a draft PR with some changes and the last commit has a comment on the callback that was not supposed to be called.

PR #1778, opened 2024-03-01
[APP-2226] Appview for Related and Versions tabs

GuiFPires:
**What does this PR do?**

This PR fixes the repopulating of the Related and Versions tabs for the appview by replacing the lazyItems inside of those tabs with "normal" items, and it also hides the Related tab when the app doesn't have related editorials.

Note that on CustomScrollableTabRow the indicatorWidths list has a fixed size, if we change the number of possible tabs in the appview we should change that as well.

**Database changed?**

 No

- [ ] AppViewScreen.kt
- [ ] OtherVersionsView.kt
- [ ] RelatedContentView.kt
- [ ] CustomScrollableTabRow.kt

The tester should open several appviews and test going in and out of Related and Versions (to info for example) to see if they don't get duplicated content in the appview or content where it shouldn't be.

They should also open and close different appviews to see if the Related tab doesn't appear on some and to see if the app doesn't crash on apps that have Related content.

  Flow on how to test this or QA Tickets related to this use-case: [APP-2226](https://aptoide.atlassian.net/browse/APP-2226)

  Tickets related to this pull-request: [APP-2226](https://aptoide.atlassian.net/browse/APP-2226)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

[APP-2226]: https://aptoide.atlassian.net/browse/APP-2226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/cm/aptoide/pt/appview/OtherVersionsView.kt
+++ b/app/src/main/java/cm/aptoide/pt/appview/OtherVersionsView.kt
@@ -10,8 +10,6 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentWidth
-import androidx.compose.foundation.lazy.LazyListScope
-import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
@@ -32,13 +30,15 @@ import cm.aptoide.pt.feature_apps.presentation.AppsListUiState
 import cm.aptoide.pt.feature_apps.presentation.appVersions
 
 @Composable
-fun OtherVersionsView(packageName: String, listScope: LazyListScope?) {
+fun OtherVersionsView(packageName: String) {
   val (uiState, _) = appVersions(packageName = packageName)
-
-  listScope?.item { Box(modifier = Modifier.padding(top = 26.dp)) }
-  (uiState as? AppsListUiState.Idle)?.run {
-    listScope?.items(apps) { app ->
-      OtherVersionRow(app)
+  Column(
+    modifier = Modifier.padding(top = 24.dp)
+  ) {
+    (uiState as? AppsListUiState.Idle)?.run {
+      apps.forEach { app ->
+        OtherVersionRow(app)
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/cm/aptoide/pt/appview/OtherVersionsView.kt
```diff
@@ -32,13 +30,14 @@ import cm.aptoide.pt.feature_apps.presentation.AppsListUiState
...
-fun OtherVersionsView(packageName: String, listScope: LazyListScope?) {
+fun OtherVersionsView(packageName: String) {
   val (uiState, _) = appVersions(packageName = packageName)
-
-  listScope?.item { Box(modifier = Modifier.padding(top = 26.dp)) }
-  (uiState as? AppsListUiState.Idle)?.run {
-    listScope?.items(apps) { app ->
-      OtherVersionRow(app)
+  Column {
+    Box(modifier = Modifier.padding(top = 24.dp))
```
pedroagostinhoaptoide, line 36: This padding can be applied to the parent Column, instead of creating a new Box. An alternative would also be to use Spacer instead of Box

app/src/main/java/cm/aptoide/pt/appview/RelatedContentView.kt
```diff
@@ -40,15 +38,17 @@ import cm.aptoide.pt.feature_reactions.ui.ReactionsView
...
+
 ) {
   val editorialsMetaViewModel = relatedEditorialsCardViewModel(packageName = packageName)
   val uiState by editorialsMetaViewModel.uiState.collectAsState()
-
-  listScope?.item { Box(modifier = Modifier.padding(top = 24.dp)) }
-  listScope?.items(uiState ?: emptyList()) { editorialMeta ->
-    RelatedContentCard(editorialMeta, onRelatedContentClick)
+  Column {
+    Box(modifier = Modifier.padding(top = 24.dp))
```
pedroagostinhoaptoide, line 46: This padding can be applied to the parent `Column`, instead of creating a new `Box`. An alternative would also be to use `Spacer` instead of `Box`

app/src/main/java/cm/aptoide/pt/appview/RelatedContentView.kt
```diff
@@ -40,15 +38,17 @@ import cm.aptoide.pt.feature_reactions.ui.ReactionsView
...
+    (uiState ?: emptyList()).forEach { editorialMeta ->
```
pedroagostinhoaptoide, line 47: Can be replaced with `uiState?.forEach`

pedroagostinhoaptoide (approved):

PR #1787, opened 2024-03-08
[APP-2236] Appview store_name changes

GuiFPires:
**What does this PR do?**

This PR changes three main requests related to appview:
- getApp request now passes the aptoide-web instead of storeName that always related to the MARKET_NAME - apps (or null in case of the wallet) for the store_name parameter
- getAppVersions request now passes the aptoide-web for the store_name parameter
- getRelatedArticlesMeta (not exactly an API but uses the curation API) now passes the aptoide-web instead of storeName that always related to the MARKET_NAME - apps

P.S. I didn't change apps/getRecommended/ and apps/getPackages because they don't seem to be used (yet) and I don't know what's the real behaviour we want from it
P.S2. I removed the exception for `com.appcoins.wallet` in the getApp because I believe with aptoide-web we won't have problems since the wallet is always there too.

(If you feel it makes sense to change one of both APIs I didn't change or to keep the exception for the wallet please tell me so I can amend the commit with the proper changes - and also explain me why so I don't have to repeat the same mistake twice)

P.S3. It was discussed today that the use cases for apps that are not in apps or catappult will be dealt with in the future.

**Database changed?**

No

- [ ] AptoideAppsRepository.kt
- [ ] AptoideEditorialRepository.kt
- [ ] RepositoryModule.kt

Open appviews, their related, versions and info tab and see if the behavior is the supposed. Test this from home (games, apps and bonus) as well as from search - Give special attention to AppCoins Wallet as well.

  Flow on how to test this or QA Tickets related to this use-case: [APP-2236](https://aptoide.atlassian.net/browse/APP-2236)

  Tickets related to this pull-request: [APP-2236](https://aptoide.atlassian.net/browse/APP-2236)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

### DIFF
--- a/feature_apps/src/main/java/cm/aptoide/pt/feature_apps/data/AptoideAppsRepository.kt
+++ b/feature_apps/src/main/java/cm/aptoide/pt/feature_apps/data/AptoideAppsRepository.kt
@@ -82,7 +82,7 @@ internal class AptoideAppsRepository @Inject constructor(
 
       appsRemoteDataSource.getApp(
         path = packageName,
-        storeName = if (packageName != "com.appcoins.wallet") storeName else null,
+        storeName = "aptoide-web",
         bypassCache = if (bypassCache) CacheConstants.NO_CACHE else null
       )
         .nodes.meta.data
@@ -138,7 +138,7 @@ internal class AptoideAppsRepository @Inject constructor(
       val randomAdListId = UUID.randomUUID().toString()
       appsRemoteDataSource.getAppVersionsList(
         path = packageName,
-        storeName = storeName
+        storeName = "aptoide-web"
       )
         .list
         ?.map { appJSON ->
--- a/feature_editorial/src/main/java/cm/aptoide/pt/feature_editorial/data/AptoideEditorialRepository.kt
+++ b/feature_editorial/src/main/java/cm/aptoide/pt/feature_editorial/data/AptoideEditorialRepository.kt
@@ -18,7 +18,6 @@ internal class AptoideEditorialRepository @Inject constructor(
   private val campaignRepository: CampaignRepository,
   private val campaignUrlNormalizer: CampaignUrlNormalizer,
   private val editorialRemoteDataSource: Retrofit,
-  private val storeName: String,
 ) : EditorialRepository {
 
   override suspend fun getLatestArticle(): List<ArticleMeta> =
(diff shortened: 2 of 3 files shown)


feature_apps/src/main/java/cm/aptoide/pt/feature_apps/data/AptoideAppsRepository.kt
```diff
@@ -82,7 +82,7 @@ internal class AptoideAppsRepository @Inject constructor(
 
       appsRemoteDataSource.getApp(
         path = packageName,
-        storeName = if (packageName != "com.appcoins.wallet") storeName else null,
+        storeName = "aptoide-web",
```
G0retZ, line 85: This change will break DT behaviour

feature_apps/src/main/java/cm/aptoide/pt/feature_apps/data/AptoideAppsRepository.kt
```diff
@@ -138,7 +138,7 @@ internal class AptoideAppsRepository @Inject constructor(
       val randomAdListId = UUID.randomUUID().toString()
       appsRemoteDataSource.getAppVersionsList(
         path = packageName,
-        storeName = storeName
+        storeName = "aptoide-web"
```
G0retZ, line 141: This change will break DT behaviour

feature_editorial/src/main/java/cm/aptoide/pt/feature_editorial/data/AptoideEditorialRepository.kt
```diff
@@ -46,7 +45,7 @@ internal class AptoideEditorialRepository @Inject constructor(
   }
 
   override suspend fun getRelatedArticlesMeta(packageName: String): List<ArticleMeta> =
-    editorialRemoteDataSource.getRelatedArticlesMeta(packageName, storeName)
+    editorialRemoteDataSource.getRelatedArticlesMeta(packageName, "aptoide-web")
```
G0retZ, line 48: This change will break DT behaviour

feature_editorial/src/main/java/cm/aptoide/pt/feature_editorial/di/RepositoryModule.kt
```diff
@@ -23,11 +22,9 @@ internal object RepositoryModule {
...
     @RetrofitV7ActionItem retrofit: Retrofit,
-    @StoreName storeName: String
   ): EditorialRepository = AptoideEditorialRepository(
     campaignRepository,
     campaignUrlNormalizer,
-    retrofit.create(AptoideEditorialRepository.Retrofit::class.java),
-    storeName
+    retrofit.create(AptoideEditorialRepository.Retrofit::class.java)
   )
 }
```
G0retZ, line 30: new line is missing :-)

G0retZ (changes requested):

PR #2153, opened 2024-12-17
[AND-215] Add Fetch package downloader

pedroagostinhoaptoide:
**What does this PR do?**

   - Adds a package downloader that downloads files using the Fetch library.

**Database changed?**

   No

See by commit.

  Flow on how to test this or QA Tickets related to this use-case: [AND-215](https://aptoide.atlassian.net/browse/AND-215)

  Tickets related to this pull-request: [AND-215](https://aptoide.atlassian.net/browse/AND-215)

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

[AND-215]: https://aptoide.atlassian.net/browse/AND-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AND-215]: https://aptoide.atlassian.net/browse/AND-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
+++ b/aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
@@ -1,0 +1,211 @@
+package cm.aptoide.pt.installer.fetch
+
+import android.content.Context
+import androidx.core.net.toFile
+import androidx.core.net.toUri
+import cm.aptoide.pt.aptoide_network.di.DownloadsOKHttp
+import cm.aptoide.pt.extensions.checkMd5
+import cm.aptoide.pt.install_manager.AbortException
+import cm.aptoide.pt.install_manager.dto.InstallPackageInfo
+import cm.aptoide.pt.install_manager.dto.InstallationFile
+import cm.aptoide.pt.install_manager.dto.hasObb
+import cm.aptoide.pt.install_manager.workers.PackageDownloader
+import cm.aptoide.pt.installer.InstallerWorker
+import cm.aptoide.pt.installer.di.DownloadsPath
+import cm.aptoide.pt.installer.platform.InstallPermissions
+import com.tonyodev.fetch2.Download
+import com.tonyodev.fetch2.Error
+import com.tonyodev.fetch2.Fetch
+import com.tonyodev.fetch2.FetchConfiguration
+import com.tonyodev.fetch2.NetworkType
+import com.tonyodev.fetch2.Priority
+import com.tonyodev.fetch2.Request
+import com.tonyodev.fetch2core.Downloader
+import com.tonyodev.fetch2okhttp.OkHttpDownloader
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onCompletion
+import okhttp3.OkHttpClient
+import java.io.File
+import javax.inject.Inject
+import kotlin.coroutines.coroutineContext
+
+class FetchDownloader @Inject constructor(
+  @ApplicationContext private val context: Context,
+  private val installPermissions: InstallPermissions,
+  @DownloadsPath private val downloadsPath: File,
+  @DownloadsOKHttp private val okHttpClient: OkHttpClient
+) : PackageDownloader {
+
+  companion object {
+    const val CONCURRENT_DOWNLOAD_LIMIT = 5
+    const val RETRY_TIMES = 3
+    const val VERSION_CODE = "versioncode"
+    const val PACKAGE = "package"
+    const val FILE_TYPE = "fileType"
+  }
+
+  private val fetch: Fetch
+  private val contentHashes = mutableMapOf<Int, String>()
+
+  init {
+    Fetch.setDefaultInstanceConfiguration(
+      FetchConfiguration.Builder(context)
+        .setDownloadConcurrentLimit(CONCURRENT_DOWNLOAD_LIMIT)
+        .setAutoRetryMaxAttempts(RETRY_TIMES)
+        .setProgressReportingInterval(1000L)
+        .enableHashCheck(true)
+        .setHttpDownloader(
+          object : OkHttpDownloader(
+            okHttpClient,
+            fileDownloaderType = Downloader.FileDownloaderType.PARALLEL
+          ) {
+            override fun verifyContentHash(
+              request: Downloader.ServerRequest,
+              hash: String
+            ): Boolean {
+              return request.fileUri.toFile().checkMd5(contentHashes[request.id] ?: hash)
+            }
+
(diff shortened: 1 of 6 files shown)


aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+          ) {
+            downloadsProgresses.replace(download.id, download.downloaded)
+            trySend(downloadsProgresses.values.sum() / totalSize)
+          }
+        }
+
+        fetch.addListener(listener)
+
+        awaitClose {
+          fetch.removeListener(listener)
```
G0retZ, line 191: Will this be called if user canceled the download?

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+        close(e)
+      }
+    }
+      .distinctUntilChanged()
+      .onCompletion {
+        if (it != null) {
+          if (it is CancellationException) {
+            fetch.cancelGroup(packageName.asGroupId())
+          } else {
+            fetch.deleteGroup(packageName.asGroupId())
```
G0retZ, line 203: what is the difference between `delete` and `remove`? 🤔

G0retZ (changes requested):

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+        trySend(0.0)
+
+        if (downloads.isEmpty()) {
+          trySend(1.0)
+          close()
+        }
+
+        fetch.enqueue(requests) { enqueuedRequests ->
+          enqueuedRequests.find { it.second != Error.NONE }?.let {
+            close(it.second.throwable ?: IllegalStateException("Error enqueuing requests"))
```
phillipsgoncalves, line 155: should we close if there is an error ?
How do they know if there are "repeated files", what is the id ?
Couldnt we ignore the repeated one instead of giving error ?

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+          }
+        }
+
+        val listener = object : FetchFileDownloadListener {
```
phillipsgoncalves, line 159: Why we didnt use the cancel or remove or delete callback ? Wouldnt they be necessary? I know we are handling it ourselves but i am just asking to be sure if there are more possible ways for it to be called

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+        }
+      } catch (e: Throwable) {
+        close(e)
+      }
+    }
+      .distinctUntilChanged()
+      .onCompletion {
+        if (it != null) {
+          if (it is CancellationException) {
+            fetch.cancelGroup(packageName.asGroupId())
```
phillipsgoncalves, line 201: Is the cancel going to keep the download for future restart and the delete completely remove it?

phillipsgoncalves (changes requested):

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
pedroagostinhoaptoide: `remove` - stops group downloads and removes temporary download files
`delete` - deletes all files from the group, even completed files
`cancel` - stops group downloads and does not remove anything

So delete should not be used, I'll fix this part.

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
pedroagostinhoaptoide: I will, unless the user manages to cancel before the awaitClose is reached. I'll add an additional call to `removeListener` on the catch block

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
```diff
@@ -0,0 +1,215 @@
...
+        .setHttpDownloader(
+          object : OkHttpDownloader(
+            okHttpClient,
+            fileDownloaderType = Downloader.FileDownloaderType.PARALLEL
+          ) {
+            override fun verifyContentHash(
+              request: Downloader.ServerRequest,
+              hash: String
+            ): Boolean {
+              return request.fileUri.toFile().checkMd5(contentHashes[request.id] ?: hash)
```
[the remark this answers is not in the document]
pedroagostinhoaptoide: Yes, it is handled by Fetch and returned through the `onError` callback from Fetch

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
pedroagostinhoaptoide: - If there is an error enqueuing one of the requests, then that file will not be downloaded so it is probably better to throw an error asap.
- Not 100% sure, but if the request local file uri is the same then Fetch will treat it as a duplicate/repeated file.
- The request id is generated automatically by Fetch.
- There should be no repeated ones, since each request is created based on the installation files

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
pedroagostinhoaptoide: I don't think there is any action that needs to be done after a file is removed/deleted so those callbacks are not necessary. Regarding the cancel callback, since the cancellation is only caught on the flow's completion, both the callbackFlow and listener won't be active anymore

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
pedroagostinhoaptoide: The cancel only cancels the download and does not remove any file. The delete removes all the group files, even successfully downloaded ones. The remove deletes only temporary files.

The logic on this part changed a bit and delete is not used any longer.

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
G0retZ: You could do this from the listener itself with `fetch.removeListener(this)`. Right before calling `trySend` or `close`. Just check the `isActive` flag...

G0retZ (approved):

aptoide-installer/src/main/java/cm/aptoide/pt/installer/fetch/FetchDownloader.kt
phillipsgoncalves: i thought the error while enqueuing was always associated with duplication. So that is why i was saying we could just ignore

PR #2254, opened 2025-03-05
[AND-347] Check internet connection before AutoUpdates

phillipsgoncalves:
**What does this PR do?**

This PR aims at implementing internet availability checks before performing the auto updates. 
We had the Worker restriction for the unmetered connection, but this was not enough as it only detects if it is on wifi, not if that wifi has a stable connection and we detected several internet errors on the the auto update flow.

**Database changed?**

No

- [ ] AutoUpdateWorker.kt

Change the code for the initial delay to be in 5 minutes so you have time to do the next step.
Run the app, install through adb with app id an older version of for example mobile legends. 
Check if the autoupdate ran correctly.
Test all cases, where the ping fails, so it needs to do a retry and confirm that it only does one retry.

  Tickets related to this pull-request: [AND-347](https://aptoide.atlassian.net/browse/AND-347)

**Questions:**

**Code Review Checklist**

- [ ] Codestyle
- [ ] Functional tests pass

[AND-347]: https://aptoide.atlassian.net/browse/AND-347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature_updates/src/main/java/cm/aptoide/pt/feature_updates/data/AutoUpdateWorker.kt
+++ b/feature_updates/src/main/java/cm/aptoide/pt/feature_updates/data/AutoUpdateWorker.kt
@@ -2,6 +2,7 @@ package cm.aptoide.pt.feature_updates.data
 
 import android.content.Context
 import androidx.hilt.work.HiltWorker
+import androidx.work.BackoffPolicy
 import androidx.work.Constraints
 import androidx.work.CoroutineWorker
 import androidx.work.ExistingPeriodicWorkPolicy.KEEP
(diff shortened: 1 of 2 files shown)


feature_updates/src/main/java/cm/aptoide/pt/feature_updates/data/AutoUpdateWorker.kt
```diff
@@ -40,9 +52,28 @@ class AutoUpdateWorker @AssistedInject constructor(
...
           .build()
       )
 
     fun cancel(context: Context) = WorkManager.getInstance(context).cancelUniqueWork(WORK_NAME)
   }
+
+  private fun isInternetAvailable(): Boolean {
+    return try {
+      val url = URL("https://www.aptoide.com")
+      val connection = url.openConnection() as HttpURLConnection
```
pedroagostinhoaptoide, line 68: It might be better if the request is a HEAD request, to have a faster response and avoid getting a full response that won't be used.

pedroagostinhoaptoide (changes requested):

Culurio (approved):

pedroagostinhoaptoide (approved):

PR #2548, opened 2026-05-26
Vanilla V10: add brand flavor + translate Figma surfaces

jdandrade:
## Summary
- Adds the **vanilla** brand flavor to `:app-games` (app id `cm.aptoide.pt`, Aptoide wordmark, vanilla launcher, vanilla light/dark palette, vanilla strings, vanilla Firebase project, `aptoide://` deep link, globally-scoped widget queries, vanilla-aware editorial network shape).
- Translates the V10 Figma into the shared composables behind `BuildConfig.FLAVOR_brand == "vanilla"` guards: Home/Editorial single-item fill, Categories rounded peach tiles with dark labels, Updates banner peach + 1.dp orange border + 16.dp rounded, Search sponsored banner peach card with top "Suggested for You / 🔥 Sponsored" strip and pill Install, and the search→GameGenie redirect disabled for vanilla so the new banner actually renders.
- `aptoideGames` flavor is preserved byte-for-byte; every change either lives in `src/vanilla/...` or is gated at runtime.

## Test plan
- [ ] `./gradlew :app-games:installVanillaDevDebug :app-games:installAptoideGamesDevDebug` builds both flavors cleanly.
- [ ] Vanilla home / categories / updates / search render per `.context/figma/*.png` (screenshots in `.context/vanilla-*-v1.png`).
- [ ] AG home / categories / updates / search rendering matches pre-branch HEAD.
- [ ] Vanilla search submits to Search Results (not GameGenie); AG search still redirects to GameGenie.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added brand-aware app variants (aptoideGames / vanilla) and variant-aware build tasks/output names.

* **Improvements**
  * Revamped theming to use dynamic palettes and propagate theme tokens; many UI components now adapt by brand (shapes, colors, status bar).
  * Several icons/drawables and compose call sites updated for Compose compatibility; notification and level/coin colors standardized.

* **Localization**
  * Added or updated many localized string resources across numerous locales.

* **Documentation**
  * Expanded build and app-variant guidance with verification steps.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Aptoide/aptoide-client-v8/pull/2548?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/backgrounds/LevelUpBackground.kt
```diff
@@ -190,10 +190,10 @@ fun getLevelUpBackground(level: Int = 0): ImageVector = ImageVector.Builder(
 }.build()
 
 private fun getLevelColor(level: Int) = when (level) {
-  1, 2 -> Palette.Orange150
-  3, 4 -> Palette.Blue100
-  5, 6 -> Palette.Yellow100
-  7, 8 -> Palette.Blue100
-  9, 10 -> Palette.Yellow100
-  else -> Palette.Orange150
+  1, 2 -> Color(0xFFF58932)
```
phillipsgoncalves: Why did you remove the Palette colors here and started using the hardcoded ones ?

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/play_and_earn/levels/tier/TierCoin.kt
```diff
@@ -85,10 +85,10 @@ fun getTierCoinIcon(level: Int): ImageVector = ImageVector.Builder(
 }.build()
 
 private fun Int.getCoinColors(): List<Color> = when (this) {
-  0, 1 -> listOf(Palette.Orange200, Palette.Yellow100, Palette.Orange150)
-  2, 3 -> listOf(Palette.Blue100, Palette.White, Palette.Blue200)
-  4, 5 -> listOf(Palette.Yellow200, Palette.Yellow50, Palette.Yellow100)
-  6, 7 -> listOf(Palette.Blue150, Palette.Blue50, Palette.Blue250)
-  8, 9 -> listOf(Palette.Yellow150, Palette.Yellow100, Palette.Yellow200)
+  0, 1 -> listOf(Color(0xFFC04D07), Color(0xFFFFC93E), Color(0xFFF58932))
```
phillipsgoncalves: Why did you stop using the Palette colors and started using the hardcoded ones ?

app-games/build.gradle.kts
```diff
@@ -117,9 +100,44 @@ android {
...
+      )
+      buildConfigField(
+        type = "String",
+        name = "DEEP_LINK_SCHEMA",
+        value = "\"ag://\""
+      )
+      manifestPlaceholders["deepLinkScheme"] = "ag"
+    }
+
+    create("vanilla") {
```
phillipsgoncalves: why didnt we just use the app module ?

phillipsgoncalves (commented):
Im not sure if i agree with this architecture of having vanilla as product flavour. The reason for that is that in many places we will have to check if the build flavour is vanilla or not. We will polute a lot of parts of the code with that.

If we want to follow this, maybe we should brainstorm tomorrow to try and find a solution to make it cleaner. For example, to inject modifiers and have a ModifierProvider which is implemented on both folders, giving the correct modifier to certain views. This wouldn't be the best solution for readability though, that is why im saying maybe we could brainstorm this.
To add ifs almost everywhere doesnt seem too good as it ads complexity, its hard to manage and is not scalable (in case we want more products for example) . For example, if we do a ticket for aptoide games, we will have to be sure to not break vanilla, or add a safeguard for it in case it is needed, this will increase testing time.
I understand that the previous solution with a different app module would make us repeat some code but at least on each of the app modules it would be readable, configurable and isolated.

Other comments:
- reverted the usage of palette and started using colors directly on some of the ui's (i commented some of them)
- on Aptoide Games, the editors choice app icon has rounded corners, it didnt have it before.
- on Aptoide Games, trending bundle stopped working.
- also, some see more views of some bundles stoped working (trending for example) on both products.
- still missing some ui things (for example the wand of for you is green), but i think it is still a work in progress
- we will need a setting for theme only for vanilla (also check feedback flow).

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/Forward.kt
```diff
@@ -63,12 +64,21 @@ fun getForward(
       strokeLineMiter = 4.0f,
       pathFillType = EvenOdd
     ) {
-      moveTo(32.615f, 32.6153f)
-      lineTo(-0.6157f, 32.6153f)
-      lineTo(-0.6157f, -0.6155f)
-      lineTo(32.615f, -0.6155f)
-      lineTo(32.615f, 32.6153f)
-      close()
+      if (BuildConfig.FLAVOR_brand == "vanilla") {
```
phillipsgoncalves: i think it would be much better to put this in another file inside the specific vanilla folder

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/LeftArrow.kt
```diff
@@ -62,12 +63,21 @@ fun getLeftArrow(color: Color, bgColor: Color): ImageVector = ImageVector.Builde
       strokeLineMiter = 1.0f,
       pathFillType = PathFillType.EvenOdd
     ) {
-      moveTo(-0.615236f, 32.6154f)
-      lineTo(32.6155f, 32.6154f)
-      lineTo(32.6155f, -0.61537f)
-      lineTo(-0.615231f, -0.615373f)
-      lineTo(-0.615236f, 32.6154f)
-      close()
+      if (BuildConfig.FLAVOR_brand == "vanilla") {
```
phillipsgoncalves: same as above

app-games/src/main/java/com/aptoide/android/aptoidegames/feature_promotional/AptoidePromotionalFeatureGraphicImage.kt
```diff
@@ -56,13 +57,41 @@ fun AptoidePromotionalFeatureGraphicImage(
...
+          horizontalAlignment = Alignment.CenterHorizontally,
+          modifier = Modifier
+            .align(Alignment.TopEnd)
+            .padding(top = 8.dp, end = 8.dp)
+            .clip(RoundedCornerShape(16.dp))
+            .background(Color(0xFFFFFFFF))
+            .padding(horizontal = 12.dp, vertical = 6.dp)
+        ) {
+          Text(
+            text = "up to",
```
phillipsgoncalves: hardcoded text

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/backgrounds/LevelUpBackground.kt
jdandrade: Good catch. These were hardcoded because `Palette` became a `@Composable` facade (LocalPalette.current) when the Vanilla flavor was added, and `getLevelColor` runs inside the non-@Composable `ImageVector.Builder` lambda, so `Palette.*` can't be called there. Switched to the plain `DarkPalette` object — same hex values, semantic tokens restored, and these are AG-only P&E visuals so no visual change. Fixed in 84061d1580.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/play_and_earn/levels/tier/TierCoin.kt
jdandrade: Same as LevelUpBackground — `Palette` is now a @Composable getter and `getCoinColors` is called inside the non-@Composable vector builder. Using `DarkPalette.*` (identical hex). Fixed in 84061d1580.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/Forward.kt
jdandrade: Done — moved the brand-specific outline into `PathBuilder` extensions (`forwardOutline()`/`leftArrowOutline()`) defined per source set (src/vanilla = circle, src/aptoideGames = square). The shared vector just calls them now, no BuildConfig branch. Verified both flavors render identically on-device. Fixed in c96c704a08.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/LeftArrow.kt
jdandrade: Same fix as Forward — outline moved to per-flavor `leftArrowOutline()` in src/vanilla and src/aptoideGames. c96c704a08.

app-games/src/main/java/com/aptoide/android/aptoidegames/feature_promotional/AptoidePromotionalFeatureGraphicImage.kt
jdandrade: You're right — moved to string resources (`promotional_bonus_up_to` / `_amount` / `_label`). Fixed in 2091be2788.

phillipsgoncalves:
There were more areas where claude changed the palette colors to hardcoded colors, maybe you should review them as well.
Also to use darkpalette seems just a workaround around the fact of the pallete being composable now meaning it cant be used in some places. This is the root problem, maybe we should find a solution for this, because someday maybe we will want to put darkpalette as composable for example.

phillipsgoncalves:
Also, we have been using the if (buildflavour == vanilla) condition on several places, i understand it is for the simplicity for this release, but on the icons we can just put them on the seperate folder specific for the product. You already changed it for one of the icons, but i think there were more. Maybe you should also ask claude to review the commits for this as well.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/LeftArrow.kt
jdandrade: Good call to flag this — there were two more icons with the exact same inline `FLAVOR_brand` branch that weren't called out: `BonusIconRight` and `BonusIconLeft` (they also gated the colors, not just the outline). Gave them the same treatment — outline + color resolution moved into per-source-set `BonusIconBrand.kt` helpers. `drawables/` is now free of inline `FLAVOR_brand` checks. e47dcedc9a

phillipsgoncalves:
you applied the palette usage fix for a certain commit, but there are several changes from the previous commits that still have hardcoded colors, can you reapply that to those ?

phillipsgoncalves:
Comments from the tests:
- app icons on bundles have rounded corners but the progress is rect. Is this supposed to be like this?
- appcoins gift is rounded, is it supposed to be like that? i feel like if it was a square would look more like an annotation
- categories: "zh tw games" ?
- Settings 1: hardware specs when we copy it goes with AG version and AG: apps (i assume it is the store)
- Settings 2: send feedback email goes to ag email. Also the text in it has the same problem as above.
- some bundles dont have see more. This is because of a rule you changed regarding the action. But this also means that no bundle has see more view. Only the carlos bundles have see more. Im not sure if this is the intended behaviour. For example editors choice and trending dont have see more.
- Gamegenie migration: the chess icon on the "or tap your favourite game and i will help you master it" is green.
- we should have a dev version of the icon for vanilla
- Really important: It is still using the AG user agent on all requests. We should provide a different user agent for vanilla. We can have a user agent provider on each folder (vanilla folder or ag folder) that would provide the correct thing.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/backgrounds/LevelUpBackground.kt
jdandrade: Addressed in 4828e13a96. The DarkPalette fallback is gone here — color resolution is hoisted to the composable call-site: a new \`@Composable levelUpBackgroundColor(level)\` resolves via \`Palette.X\` and the resolved \`Color\` is passed into the non-composable \`getLevelUpBackground(levelColor)\` builder, so the builder no longer touches the palette at all.

app-games/src/main/java/com/aptoide/android/aptoidegames/drawables/icons/play_and_earn/levels/tier/TierCoin.kt
jdandrade: Same treatment in 4828e13a96 — \`getCoinColors()\`/\`DarkPalette\` removed. A new \`@Composable tierCoinColors(level)\` resolves via \`Palette.X\` at the \`TierBadge\` call-site and passes the \`List<Color>\` into \`getTierCoinIcon(colors)\`. \`grep DarkPalette app-games/src/main\` is now empty.

jdandrade:
@phillipsgoncalves re: the remaining hardcoded colors + the DarkPalette workaround — both addressed in 4828e13a96:

- **Reapplied across all the PR’s commits:** every hardcoded `Color(0x…)` literal this PR introduced now routes through a single documented source of truth, `theme/FixedColors.kt` (`Dark` / `White` / `Scrim` / `VanillaOrange` / `VanillaGiftGold`).
- **Why these are not just `Palette.X`:** they are intentionally *theme-invariant*. `Palette.Black`/`Palette.White` **invert** in Vanilla’s light theme (Black→white), so reverting would turn dark-on-orange button text and icon outlines white-on-orange in light mode — verified on-device. `FixedColors` keeps them fixed and documents exactly that.
- **DarkPalette root-cause:** the `DarkPalette` fallback in the P&E vector drawables is removed. Color resolution is hoisted to the composable call-sites (`levelUpBackgroundColor` / `tierCoinColors` using `Palette.X`) and passed into the non-composable builders. `grep DarkPalette app-games/src/main` is now empty.

One thing left deliberately out of scope: the `*NotificationBuilder.kt` files still use raw `Int` colors (`0x…toInt()`) — those run outside Compose entirely (no Palette/CompositionLocal) and additionally hardcode AG lime on Vanilla, so they belong to a separate brand fix rather than `FixedColors`.

jdandrade:
@phillipsgoncalves re: the User-Agent point (“Really important: still using the AG user agent”) — fixed in 206a990abe.

Extracted the brand token into a per-source-set constant, `network/UserAgentBrand.kt` (`AptoideGames` vs `Aptoide`), exactly the per-folder provider you suggested and matching the existing `WidgetsConfig` pattern; the shared `AptoideGetHeaders` just references it. Verified on-device:

- Vanilla → `Aptoide/<version> (…; cm.aptoide.pt; …)` — identical to the production `:app` client
- Aptoide Games → `AptoideGames/<version> (…)` — unchanged

The rest of the test-findings list in that comment (rounded icons/progress, categories naming, settings AG strings, GameGenie chess icon, dev icon, etc.) is being tracked separately.

jdandrade:
@phillipsgoncalves re: the GameGenie green elements — fixed in ec9f157ecc.

The chess icon ("Or tap your fav game…") plus two more lime-green spots (the "AI" toolbar badge pixels and the chat send icon) were hardcoded to AG lime, so they stayed green on Vanilla. They now go through the brand-aware `Palette.Primary` (lime in AG, orange in Vanilla), using the same ImageVector-takes-a-Color pattern as the rest of the module. The multi-color `gamegenie_ai_toolbar_icon.xml` was converted to a Compose `ImageVector` so its pixels can use `Palette.Primary` (white pill / dark "AI" via `FixedColors`); the unused XML was deleted.

Verified on-device: Vanilla now shows orange in all three spots, Aptoide Games is visually unchanged (lime).

PR #2579, opened 2026-08-28
[AND-869] Fall back to the Play details overlay when the inline install deep link is rejected

jdandrade:
## Problem

Inline installs are broken in production on the Play build: tapping Install does nothing for Block Blast / Brawl Stars despite valid catalog tokens.

Root cause (verified on-device, Play Store 52.6 & 52.8): Finsky rejects the documented Catalog Access `/d` deep link **before token validation**, logging `PITH: called from wrong URI` and killing the transparent half-sheet activity in ~500ms with no UI and no error signal to the caller. The rejection only happens for the **enrolled caller identity** — unenrolled callers get a working sheet from the same URI, so this is a Google-side regression tied to our 3PAS registration (evidence package sent to Google support, answer pending).

## Fix: fallback ladder

The only signal available is the install UI closing, so closing time now tells rejection apart from user cancellation — an instant close (<2s) advances the ladder instead of cancelling:

1. **`/d` + catalog token** — documented 3PAS route, kept primary so it self-heals if Google fixes it
2. **`store/apps/details` overlay, no token** — Play's public referral route, verified working end-to-end (half-sheet → install attributed to `com.android.vending`). Deliberately token-less: this route hard-aborts on a token it cannot validate
3. **Regular Aptoide download+install, continued automatically** — the user never saw any of the rejected stages, so their tap intent stands

Behavior changes:
- A user-dismissed sheet still cancels, but no longer aborts inline installs for the session — only an invisibly rejected ladder does (previously any cancel aborted, which would starve the fallback on retries)
- Fallback overlay installs report a distinct `google_play_overlay` install method — a referral differs from a Catalog Access distribution commercially

## Verification (on-device, gplay dev build)

- Full ladder exercised end-to-end with logs at every hop: primary launch → instant-close detected → fallback details sheet shown → second instant-close → regular download auto-started (no extra tap)
- Late close (user dismissal) correctly routed to cancel, no session abort
- Token-less apps unchanged: `no catalog token -> regular install path`, ladder never starts
- Prod-conditions caveat: the enrolled-caller `/d` rejection only reproduces with the Play-signed prod package, so the primary-stage rejection was stood in for by fast manual dismissal (threshold temporarily raised during the test); final confirmation on the next Play prod release

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app-games/src/gplay/java/com/aptoide/android/aptoidegames/installer/gplay/AptoideCatalogTokenRepository.kt
+++ b/app-games/src/gplay/java/com/aptoide/android/aptoidegames/installer/gplay/AptoideCatalogTokenRepository.kt
@@ -13,7 +13,10 @@ class AptoideCatalogTokenRepository @Inject constructor(
   private val playInlineConfigApi: PlayInlineConfigApi,
 ) : CatalogTokenRepository {
 
-  override suspend fun getCatalogToken(packageName: String): ByteArray? = runCatching {
+  // Passed to Play as-is: Google support confirmed catalog_token is a String extra
+  // (their docs wrongly showed byte[] until 2026-08; a byte[] extra makes Finsky read
+  // null and kill the half-sheet with a misleading "PITH: called from wrong URI" log)
+  override suspend fun getCatalogToken(packageName: String): String? = runCatching {
     withTimeout(TOKEN_FETCH_TIMEOUT_MILLIS) {
       playInlineConfigApi.getPlayInlineConfig(packageName)
     }
@@ -24,9 +27,6 @@ class AptoideCatalogTokenRepository @Inject constructor(
     .getOrNull()
     ?.catalogToken
     ?.takeIf { it.isNotBlank() }
-    // TODO(inline-installs): confirm with a real export token whether the backend serves it
-    //  Base64-encoded (protobuf bytes) and needs decoding before being passed to Play
-    ?.toByteArray()
 
   private companion object {
     const val INLINE_INSTALL_TAG = "InlineInstall"
--- a/app-games/src/gplay/java/com/aptoide/android/aptoidegames/installer/gplay/CatalogTokenRepository.kt
+++ b/app-games/src/gplay/java/com/aptoide/android/aptoidegames/installer/gplay/CatalogTokenRepository.kt
@@ -10,8 +10,11 @@ import javax.inject.Inject
  */
 interface CatalogTokenRepository {
 
-  /** Returns a fresh catalog token for [packageName], or null if none is available. */
-  suspend fun getCatalogToken(packageName: String): ByteArray?
+  /**
+   * Returns a fresh catalog token for [packageName], or null if none is available.
+   * A String, passed to Play as-is - see [AptoideCatalogTokenRepository].
+   */
+  suspend fun getCatalogToken(packageName: String): String?
 }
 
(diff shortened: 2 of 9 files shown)


jdandrade:
Update after Google support's reply: their engineers determined `catalog_token` must be a **String extra, not `byte[]`** — their docs showed `byte[]` and have now been corrected. This explains the whole failure: Finsky's enrolled-caller path read our byte[] extra as a null String and killed the half-sheet with the misleading "wrong URI" log (unenrolled callers never reach token parsing, which is why they got a working sheet).

Second commit adjusts accordingly:
- `catalog_token` now passed as a String end-to-end (`CatalogTokenRepository` returns String; resolves the old token-encoding TODO)
- The details-overlay fallback stage is removed per internal decision: catalog apps use Google's documented method or Aptoide's own install path, never the public overlay (opted-out developers must not surface through us). The overlay route is reserved for the upcoming rewarded-ads flow. Ladder is now `/d` → regular install, which the on-device verification already covered (no fallback resolved → exhaustion → auto-continued regular install).

Caveat: only the Play-signed enrolled package exercises Play's strict token path, so the String fix gets its definitive validation on the next Play release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jdandrade:
Third commit adds the overlay-only distribution for **Free Fire / Free Fire MAX / Roblox** (decision: gray-zone-defensible Play referral instead of direct APK serving, apkfy-driven for web-acquired P&E users, outside the searchable catalog):

- `resolveInlineInstall` routes these three packages to `store/apps/details` with **no token** (they're not in the catalog export). Everything else keeps the Catalog Access `/d` + token route.
- **No APK fallback for them**: new `InlineInstallResolver.allowsRegularFallback` — an exhausted/failed overlay ends as canceled, never as an Aptoide download. Retries reopen the overlay.
- Their installs report `google_play_overlay` (click + completed + canceled), keeping BI separated from Catalog Access installs.
- Bugfix: inline launches now go through a `Channel` (exactly-once) instead of a `SharedFlow` — the apkfy screen composes two install views per app, which double-launched the Play sheet.

Verified on-device (gplay prod-debug, real package name): apkfy screen for Free Fire → details overlay → 392 MB installed by `com.android.vending` **without the install-unknown-apps permission granted**, apkfy screen flips to Open; Brawl Stars regression still routes `/d` + token with the Catalog Access sheet.

Note for QA: the apkfy install-permission prompt is unnecessary for overlay-only titles (Play does the installing) — candidate follow-up to skip it for them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jdandrade:
Fourth commit fixes a gap the apkfy testing exposed: **install-start was never reported for diverted installs**. The regular path fires `onInstallStarted` when the constraints resolver resolves — a point Play inline/overlay installs never reach — so screens reacting to install start missed Play-side installs entirely. Most visibly, apkfy's **Play & Earn reward routing never triggered** for Free Fire installed through the details overlay, which is exactly the web-acquired P&E flow. `onInlineInstallLaunched` now reports install start as well.

QA note: please confirm the apkfy reward screen appears with a live P&E reward/campaign attached — on-device verification here covered the callback wiring, but the reward content itself depends on campaign state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

